PR #1, opened 2014-10-27
Make auth_socket more flexible.

dveeden:
Now supports "AS '<user>'" so it doesn't have to be a 1-on-1 mapping.

Example:
CREATE USER 'foo'@'localhost IDENTIFIED WITH auth_socket AS 'myuser';


roel666:
Thanks

dveeden:
The MySQL Bug ID for this:
http://bugs.mysql.com/74586

MyDanny:
Thank you. Unfortunately, we can't accept pull requests on github. I see you've already submitted your patch via the bug tracker, so I'm closing this pull request.

PR #3, opened 2015-01-25
ssl.cmake: fix version regex for OpenSSL 1.0.2.

MikeMcQuaid:
OpenSSL 1.0.2 has whitespace between the `#` and `define` so make the regex more liberal to match it accordingly.

Should be backported to 5.6, 5.5 branches too.

Thanks!

CC https://github.com/Homebrew/homebrew/issues/36205 @DomT4


MikeMcQuaid:
Oh, closing this as I see you're not accepting pull requests here (yet :wink:).

DomT4:
http://bugs.mysql.com/bug.php?id=75623

PR #4, opened 2015-02-04
Write use statement to every slowlog entry

emate:
Due to nature of slowlogs, it's hard to parse them in efficient way. Allowing to write 'use' statement to every slowlog entry greatly improve ease of parsing logs with fixed number of lines without causing any impact on system (i/o footprint is unnoticeable).


morgo:
Hi Marcin, thank you for the contribution!

We are not currently accepting pull requests from GitHub. May I suggest filing a bug report on bugs.mysql.com?

(I would also be happy to do so on your behalf.)

emate:
Hi Morgan, 

thanks for reply. I already did that, http://bugs.mysql.com/?id=75775.

morgo:
Thank you!  I am going to go ahead and close this pull request.

PR #7, opened 2015-05-08
Bug#68925: Compatibility issue with mysql history ("\040" instead of …

nirbhayc:
…space)

Problem: libedit encodes whitespace characters before dumping them to history file.

Fix: Be readline-friendly, do not encode whitespace characters.


nirbhayc:
Hello!
I cannot sign OCA. But since the fix is seemingly trivial (oneliner), I have added the patch as suggestion to Bug#68925.

PR #8, opened 2015-05-18
Update echo.c

dataScientist420:
Modifications: 
- adding an empty space for the first fprintf 
- removing the if statement 

Purpose: eliminate argc executions of the if statement and argc-1 calls of fprintf

Improvement: the produced machine code will have less instructions and less function calls (overhead)


dataScientist420:
Hi,
Thanks

PR #10, opened 2015-06-01
Add _login and _hostname as connection attribute.

dveeden:
https://bugs.mysql.com/bug.php?id=77190


dveeden:
This works on UNIX, didn't test on Windows

dveeden:

PR #11, opened 2015-06-07
Use unicode to draw borders

dveeden:
This uses unicode to draw the table borders

screenshot from 2015-06-07 23-52-13

### DIFF
--- a/client/mysql.cc
+++ b/client/mysql.cc
@@ -3840,7 +3840,7 @@ print_table_data(MYSQL_RES *result)
       return;
     mysql_field_seek(result,0);
   }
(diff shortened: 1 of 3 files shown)


bazzilic:
What problem does this solve?

dveeden:
The solid lines look better that the dashed lines.

This could be improved further by using different characters for the corners etc to get rid of the 'spikes' on the corners.

bazzilic:
What I'm saying is that as this is supposed to run on multiple platforms, there might be numerous problems with console hosts that don't fully support unicode. That seriously outweighs the benefit of having a slightly better looking output.

dveeden:

PR #13, opened 2015-06-23
Bug #76953: Many IPv6-specific tests fail due to missing IPv6 support

akopytov:
Fixed a number of IPv6-specific tests so that checks for IPv6 work as
intended: skip test when IPv6 support is unavailable instead of failing
with an error on server startup.


akopytov:

akopytov:

PR #14, opened 2015-06-23
Add RESET MASTER TO x to allow specification of binlog file nr

dveeden:
[Bug #77438](https://bugs.mysql.com/bug.php?id=77438)  Extend RESET MASTER to allow specifying next binary log.


dveeden:

PR #17, opened 2015-06-25
SUSE patches we keep against 5.6

scarabeusiv:
Hello, this is bunch of patches we keep as upstreamable in our mysql-community-server 5.6 brach.

To ease everything up oracle/mysql-project can license and do what ever they want with the patches provided by all SUSE employees.


scarabeusiv:
The oca seems to already contain "SUSE Linux Products GmbH" is that enough or you really need individual employee signatures?

scarabeusiv:

PR #18, opened 2015-06-29
Bug #74891: default cmake excludes utf8_5624_1 charset, required for main.ctype_ldml

akopytov:
Fix scan_one_character() to work correctly on architectures where char is
unsigned by default.


akopytov:

PR #20, opened 2015-07-01
Bug #77577: setup_timers initialization assumes CYCLE timer is always available

akopytov:
Test if the CYCLE timer is actually implemented when using it as the
default timer for wait events. If CYCLE is not available, fall back to
the nanosecond timer (or even less resolution timers, if nanosecond is
also unavailable), similar to how other event timing defaults are
adjusted.


akopytov:

PR #22, opened 2015-07-11
Update README

dveeden:
- Rename to README.md so it will be rendered on Github
- Use HTTPS for all links
- Use markdown for formatting
- Add note about OCA/Contributing
- Add more links (e.g. sections of the manual which are mentioned)


dveeden:

PR #23, opened 2015-07-16
Bug #76927: Duplicate UK values in READ-COMMITTED (again)

akopytov:
The problem was with `lock_rec_inherit_to_gap()` assuming that only
shared record locks can be set by a consistency constraint and thus,
only that lock type should be inherited in the gap mode when an
index record is deleted.

However, `REPLACE`, `LOAD DATA REPLACE` and `INSERT ON DUPLICATE KEY UPDATE`
set exclusive record locks. When a record had been deleted by purge, all
record locks acquired by in-progress concurrent statements of those
types were incorrectly released by `lock_rec_inherit_to_gap()` which might
lead to a UK constraint violation.

Fixed by changing `lock_rec_inherit_to_gap()` to take `trx_t::duplicates`
into account. In case it has a non-zero value (i.e. `TRX_DUP_IGNORE` or
`TRX_DUP_REPLACE`), exclusive rather than shared locks are now inherited
as gap locks.

### DIFF
--- a/mysql-test/suite/innodb/r/bug76927.result
+++ b/mysql-test/suite/innodb/r/bug76927.result
@@ -1,0 +1,39 @@
+CREATE TABLE t1 (
+a INT NOT NULL,
+b INT NOT NULL,
+PRIMARY KEY(b),
+UNIQUE KEY(a))
+ENGINE=INNODB;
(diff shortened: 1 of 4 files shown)


akopytov:

PR #24, opened 2015-07-27
Bug #72259 - Filesystem read-only detector

elshadaghazade:
Hello, I'm developer from Baku, Azerbaijan. I became an OCA member next week and i feel good :) I will try to write many usefull codes to develope MySQL in the future. Now I want to write about these changes. My friend found that when the file system is read-only, MySQL does not show exact error message about it. He reported this bug - http://bugs.mysql.com/bug.php?id=72259 . And I decided to fix this bug myself. I wrote a function - int is_filesystem_read_only(char const\* path) - this function detects if filesystem is read-only and return an integer. In mysqld_main function I catch --datadir option, get it's value and path to my new function. If file system is read-only I send an error message like this "Filesystem ([datadir path]) is read-only. Thats all :) I hope you will appreciate my first code for MySQL. Thanks a lot

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -21,6 +21,8 @@
 #include <functional>
 #include <list>
 #include <set>
+#include <unistd.h> // this library will help us to detect file system status
+#include <string.h>
 #include <string>
 
 #include <fenv.h>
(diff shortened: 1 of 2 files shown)


elshadaghazade:

PR #27, opened 2015-09-16
Evict table metric (part 2)

dveeden:
This is a follow-up for https://bugs.mysql.com/bug.php?id=78190 (Bug #78190)


dveeden:

mysql-admin:
Thanks for the contribution but the bug 78190 was already fixed and is closed

PR #29, opened 2015-10-23
Bug#78932: mysqldump --routines misquotes database names

FGasper:


FGasper:
It would be great to get a backport of this to as many versions are supported … the breakage from this is pretty debilitating when you want a full backup of everything.

FGasper:
I just sent in the OCA.

I didn’t see a place to indicate which issue I was submitting the form 
for, however..?

-FG

mysql-admin:
Hi, 
Once we process the OCA and enter it into the system you will get another notice to confirm this contribution is under the OCA and then it will be associated with bug 78932

Thanks again for your contribution
==Omer

FGasper:

Sent from my android device.

-----Original Message-----

PR #30, opened 2015-10-24
Bug #46053 GROUPING() function for ROLLUP 

jimmybupt:
Hi, I'm Zhe Dong (OCA). I added a native grouping() function as described in bug #46053.

Intended behavior: 
The grouping() function returns integer 1 if and only if its argument is a single column reference and the current row is a subtotal of the input column. It returns 0 in all other circumstance.


dveeden:
I've added it here: https://dveeden.github.io/modern-sql-in-mysql/ 
Is this a SQL:1999 feature or a SQL:2003 feature?

jimmybupt:

jimmybupt:
To dveeden,
Sorry I'm not able to answer your question. I cannot get access to any ISO standard file. All I can get is a committee draft dated 2013.

PR #31, opened 2015-11-05
Bug #79144: No hardware CRC32 implementation for AArch64

akopytov:
Add support for autodetection and usage of hardware CRC32 instructions on the AArch64 architecture, as well as some minor code generalization via the UNIV_CRC32_HW define.


akopytov:

PR #32, opened 2015-11-24
Print GTID info on STOP/START SLAVE

dveeden:
Bug #71565  Please report full GTID state when stopping / start a slave
https://bugs.mysql.com/bug.php?id=71565

This does not (yet) print Gtid Retrieved Set.


dveeden:

PR #34, opened 2015-12-11
add super_connections option

xiezhenye:
Users with SUPER privileges can use an extra connections when connection count reaches max_connections.
But the number of the reserved connections can not be changed. In some cases, more reserved connections is required for maintenance or other reason.

http://bugs.mysql.com/79606


xiezhenye:

xiezhenye:

PR #36, opened 2015-12-13
Add parser which indexes words in Unicode NFC normalized form.

dveeden:
When the text contains both normalized and non-normalized words
(e.g combining characters for e+" instead of e). Then this parser
 will index the normalized form and will return results for both if
the search key is normalized.

This is proof-of-concept code and is not yet production ready.


dveeden:

PR #38, opened 2015-12-21
innochecksum optimization: crc32 calculated already

grooverdan:
If curr_algo == SRV_CHECKSUM_ALGORITHM_CRC32 then calc already contains
the result we want so lets use that.


grooverdan:

grooverdan:

PR #39, opened 2015-12-21
Check algorithm=innodb on crc32 checksum mismatch before crc32(big endian)

grooverdan:
When a page has had its checksum fail and the algorithm = CRC32,
its most likely because we've updated from 5.6 where the default
was algorithm=innodb. We should test that algorithm innodb on the
page passes before testing the crc32 in big endian way (
page_zip_calc_checksum(..., true)) as its more likely to be true.


grooverdan:

PR #40, opened 2015-12-21
Bug #79487: Provide non-Windows implementations for LF_BACKOFF

akopytov:
Add LF_BACKOFF implementations for GCC and Sun Studio + some minor code
unification with InnoDB.

This obviously needs further unification with InnoDB code (in particular, ut_delay()) and the contributed patch for http://bugs.mysql.com/bug.php?id=74832.

But I understand this patch will not be merged as is anyway, so I tried to make it as isolated as possible, but avoid code duplication with InnoDB at the same time.

### DIFF
--- a/include/atomic/generic-msvc.h
+++ b/include/atomic/generic-msvc.h
@@ -106,30 +106,4 @@ static inline void my_atomic_storeptr(void * volatile *a, void *v)
   (void)InterlockedExchangePointer(a, v);
 }
 
-
-/*
-  my_yield_processor (equivalent of x86 PAUSE instruction) should be used
-  to improve performance on hyperthreaded CPUs. Intel recommends to use it in
-  spin loops also on non-HT machines to reduce power consumption (see e.g
-  http://softwarecommunity.intel.com/articles/eng/2004.htm)
-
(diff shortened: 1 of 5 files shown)


ghost:
See also: http://bugs.mysql.com/bug.php?id=72755
and http://bugs.mysql.com/bug.php?id=72805 which was probably incorrectly closed - as everything got kind of messy around some things going in/out of InnoDB and generic mysys code.

My comment would be that perhaps the UT_LOW_PRIORITY_CPU() and UT_RESUME_PRIORITY_CPU() macros should be called in my_yield_processor() rather than trusting the caller to also call them. Without these, the code generated for POWER in lf_hash and lf_alloc-pin will not do exactly what is intended - it'll delay, but it won't relax the CPU thread to give more core resources to other threads executing in the core.

Then there's the issue of the rest of the code in lf_hash... which I'm not convinced is correct.

akopytov:
Thanks, @stewart-ibm. I moved  UT_LOW_PRIORITY_CPU() and UT_RESUME_PRIORITY_CPU() code from InnoDB to my_atomic.h and used it in my_yield_processor().

akopytov:

PR #44, opened 2015-12-30
Bug #79378: buf_block_align() makes incorrect assumptions about chunk size

akopytov:
Code in buf_block_align() depended on srv_buf_pool_chunk_unit to find
the chunk corresponding to a given pointer. The problem was that
srv_buf_pool_chunk_unit does not necessarily reflect the real chunk size
in bytes as explained in the bug report.

Fix the algorithm buf_block_align() to not depend on any specific chunk
size, but use std::map properly to find the corresponding element in the
chunk map.


akopytov:

PR #45, opened 2015-12-30
Bug #79379: Inconsistent innodb_buffer_pool_size handling

akopytov:
Remove code reserving extra bytes for the allocated memory block in
buf_chunk_init(). Those extra bytes break the alignment even if the
originally requested chunk is already properly aligned on all
levels. That in turn results in all kinds of inconsistencies with large
OS pages in setting innodb_buffer_pool_size both statically and
dynamically.

### DIFF
--- a/mysql-test/suite/innodb/t/innodb.test
+++ b/mysql-test/suite/innodb/t/innodb.test
@@ -1394,7 +1394,7 @@ drop table t1;
 # Test for testable InnoDB status variables. This test
 # uses previous ones(pages_created, rows_deleted, ...).
 --disable_warnings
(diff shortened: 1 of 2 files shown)


akopytov:

manuelkiessling:
Is this still in the queue and not merged for 5.7.19, and thus http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/MySQL.KnownIssuesAndLimitations.html still applies?

akopytov:
@manuelkiessling Right, this has not been merged and is currently not available in any release series, not even in 8.0.

manuelkiessling:
Alright, thanks for the update!

PR #46, opened 2016-01-19
Bug #80058: Free creation_ctx in sp_load_for_information_schema where create_string fails

grooverdan:
Found during static analysis:

cppcheck --force -DDBUG_VOID_RETURN=return  -"DDBUG_RETURN(a)=return a" sql/sp.cc
[sql/sp.cc:2282]: (error) Memory leak: creation_ctx

Also exists in 5.6 and 5.7


grooverdan:

PR #47, opened 2016-01-19
Bug #80059: Correct cppcheck warning that could occur when all elements are <= 1

grooverdan:
cppcheck -DDBUG_VOID_RETURN=return  -"DDBUG_RETURN(a)=return a"  mysys/testhash.c

[mysys/testhash.c:172]: (error) Array 'key1[1000]' accessed at index 1000, which is out of bounds.

same errors occur in 5.6 and 5.7


grooverdan:

PR #51, opened 2016-02-08
Add new scheduling algorithm for reducing tail latencies 

sensssz:
This branch introduces a new scheduling algorithm (Variance-Aware-Transaction-Scheduling, VATS) for the record lock manager of InnoDB. Instead of using First-Come-First-Served (FCFS), the newly introduced algorithm chooses between FCFS and Eldest-Transaction-First (ETF), which prefers older transactions over new ones. A configuration parameter (innodb_lock_schedule_algorithm) is introduced for users to choose between VATS and FCFS (the default one). Another parameter (innodb_vats_wait_lock_pct) is introduced to specify the threshold of the percentage of wait locks of choosing ETF over FCFS. We've extensively tested this algorithm in many workloads. The algorithm is very simple, and the changes are very local, but it significantly improves performance (in terms of average latency and throughput) and predictability (in terms of reduction of tail and quantile latencies) For more details, please refer to this paper http://arxiv.org/abs/1602.01871

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -282,6 +282,22 @@ static TYPELIB innodb_checksum_algorithm_typelib = {
 	NULL
 };
 
+/** Possible values of the parameter innodb_lock_schedule_algorithm */
+static const char* innodb_lock_schedule_algorithm_names[] = {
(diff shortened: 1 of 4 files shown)


sensssz:

PR #52, opened 2016-02-08
Bug #79455: Restore get_sched_indexer_t in 5.7

akopytov:
This partially reverts the fix for bug bug#20709391 "SCHED_GETCPU() IN
DEFAULT GET_RND_INDEX() COULD NOT GENERATE RANDOM COUNTERS" by
re-introducing get_sched_indexer_t on Linux and Windows and using it (if
available) as the default indexer in ib_counter_t. It also changes
InnoDB row counter updates to use the default indexer rather than
override it with MySQL trx ID.

In sync_array_get() use the counter_indexer_t explicitly, because other
ones are not random enough per comments in the fix for bug #20709391.

This in itself does not introduce any measurable performance
improvements, but is rather a groundwork for further changes.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -8601,8 +8601,7 @@ ha_innobase::index_read(
 	case DB_SUCCESS:
 		error = 0;
 		table->status = 0;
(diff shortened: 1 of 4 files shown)


akopytov:

PR #53, opened 2016-02-17
thr_lock: Only set stack size when stack size is < 64k

daxtens:
On powerpc64le with 64k pages, attempting to set stack size to 64k fails,
so the thr_lock test fails immediately.

Instead, check the current stack size, and if it's bigger, leave it unchanged.
The test program now succeeds.

Signed-off-by: Daniel Axtens daxtens@au1.ibm.com


daxtens:
Yep, PTHREAD_STACK_MIN is 131072. I will respin to test if PTHREAD_STACK_MIN < 65536 and only adjust if that's the case.

daxtens:

daxtens:
Updated per @stewart-ibm's comments.

PR #54, opened 2016-02-18
mysys: Remove alloca() from lock-free allocator

daxtens:
A crash on powerpc64le was traced to the use of alloca() in
_lf_pinbox_real_free, where the code was attempting to allocate
252kB on the stack.

_lf_pinbox_real_free is not a hotpath, and the lock-free allocator is
used only in the performance schema and mysys/waiting_threads.c (which
appears to be dead code). So simply replace alloca() with the much
safer my_malloc(): allocate on the heap.

The first patch removes alloca(), and the second patch does cleanups: 
The calculation of stack_ends_here was only used by the lock-free
allocator, and only to do aggressive sizing of an alloca() call.
The calculation is unclear and the use in alloca() caused a crash:
remove it all.

### DIFF
--- a/include/lf.h
+++ b/include/lf.h
@@ -111,13 +111,12 @@ typedef struct {
 typedef struct {
   void * volatile pin[LF_PINBOX_PINS];
   LF_PINBOX *pinbox;
-  void  **stack_ends_here;
   void  *purgatory;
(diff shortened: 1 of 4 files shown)


daxtens:

PR #56, opened 2016-03-18
bug 80808: CHECKSUM TABLE to calculate in multiple column chunks

grooverdan:
Checksum implementations contain optimizations for calculating
checksums of larger blocks of memory.

This optimization calls my_checksum on a larger block of memory
rather than calling on multiple adjacent memory as its going though
the table columns for each table row.

Note: even though current crc32 is not optimized I'm preparing patches to change this.


dveeden:
Maybe also get this to work on a per partition basis?
https://bugs.mysql.com/bug.php?id=78733

grooverdan:

PR #57, opened 2016-03-18
bug 80809: innodb.innodb fails on PPC64

grooverdan:
Because page size differences on PPC64, the innodb_buffer_pool_pages_total
value may also be 1539.

Like MariaDB bug MDEV-6872 (fix http://bazaar.launchpad.net/~maria-captains/maria/5.5/revision/4325)


grooverdan:

PR #59, opened 2016-03-23
Bug #80783: Use default buffer pool size for faster startup with --initialize

akopytov:
For performance reasons, use the default buffer pool size rather than
the one specified in configuration when the server is started with
either --initialize or --initialize-insecure options.


akopytov:

PR #61, opened 2016-03-29
MTR: record rpl_test for replication test suite

grooverdan:
This enables --skip-rpl option on mtr to actually skip replication tests.


grooverdan:

PR #62, opened 2016-03-29
cleanup in test test_service_sql_api.test_session_general_log

grooverdan:
This enables this test to be able to be run with ./mtr --repeat=2


grooverdan:

PR #63, opened 2016-03-29
mtr ctype.gb18030_binlog and main.ctype_ldmland need to be repeatable

grooverdan:
./mtr --repeat=2 on both of these tests causes failure on the second run


grooverdan:

PR #64, opened 2016-03-29
Bug #80879: Allow '.' and '+' in compiler flags when generating mysql_config

akopytov:
Change REPLACE_REGEX_OPTIONS in scripts/CMakeLists.txt to allow dots in
-mcpu, -mtune, and -march, and the plus sign in -mcpu.


akopytov:

PR #37, opened 2015-12-15
Few improvements related to CPU cache line size and padding

akopytov:
Bug #79636: CACHE_LINE_SIZE should be 128 on AArch64
Bug #79637: Hard-coded cache line size
Bug #79638: Reconcile CACHE_LINE_SIZE with CPU_LEVEL1_DCACHE_LINESIZE
Bug #79652: Suspicious padding in srv_conc_t
- changed CPU_LEVEL1_DCACHE_LINESIZE to default to 128 bytes on POWER
  and AArch64 architectures in cases when no value could be detected
  by CMake using getconf
- changed pfs_global.h to assume CPU_LEVEL1_DCACHE_LINESIZE is always
  defined (and throw a compile-time error if it's not)
- changed CACHE_LINE_SIZE definition in ut0counter.h to be an alias of
  CPU_LEVEL1_DCACHE_LINESIZE
- changed a number of hard-coded 64-byte cache line size values in the
  InnoDB code
- fixed insufficient padding for srv_conc members in srv0conc.cc

PR #66, opened 2016-04-01
Bug #79683: Few improvements related to CPU cache line size and padding:

akopytov:
Another change in the patch submitted in https://github.com/mysql/mysql-server/pull/37 based on @grooverdan's comments. Creating a new PR to update the previous contribution.


akopytov:

grooverdan:
It seems rather than the crude padding of the line cache size, or horribly frail maths (https://github.com/mysql/mysql-server/blob/5.7/include/lf.h#L74), that alignment directives could be used to do line cache padding like:

https://gist.github.com/grooverdan/19a7c8d8ac0e680ba6fcf4668962b9c4

I haven't seen an easy fallback to compilers that don't support them but is it really important on those?

akopytov:
Agree, the resulting code is easier on the eyes. Whether it needs support for exotic compilers is up to the corresponding vendors to decide.

PR #69, opened 2016-04-08
Fix for bug#80981: Make "libs-compat" stand-alone

fromdualjb:
Let the "libs-compat" RPM contain the "charsets/*" files, and let them
be stored in a path that contains the old version number.
To be sent to Oracle under OCA.


fromdualjb:

PR #67, opened 2016-04-06
correct make_pair for c++11

grooverdan:
Adding -std=c++11 to the CXX flags as follows causes one compile fault fixed as follow.

There are lots of auto_ptr deprecation warnings too that need handling at some point. Seems a straight replacement to unique_ptr doesn't solve all cases.

cmake ../mysql-5.7 -DCMAKE_C_COMPILER=${CC:-gcc} -DCMAKE_CXX_COMPILER=${CXX:-g++} -DCMAKE_C_FLAGS_RELWITHDEBINFO="-O3 -mcpu=power8 -g"  -DCMAKE_CXX_FLAGS_RELWITHDEBINFO="-O3 -mcpu=power8 -g -std=c++11" -DWITH_SSL=system -DCMAKE_INSTALL_PREFIX=/usr/local/mysql-5.7 -DDOWNLOAD_BOOST=1 -DWITH_BOOST=/home/danielgb/build_boost  -DCMAKE_C_FLAGS="-O3 -mcpu=power8 -g" -DCMAKE_CXX_FLAGS="-O3 -mcpu=power8 -g -std=c++11"

PR #71, opened 2016-05-09
do not pass type to std::make_pair()

eworm-de:
We not pass type template arguments to std::make_pair()
explicitly, and let the compiler deduce them on its own.
This fixes compiler error on GCC 6.1.1.


grooverdan:
Same as bug 81317 dup of #67 (bug 80996). Fixed for forthcoming 5.7.14 release according to bug report. I don't know which fix was implemented.

eworm-de:
Oh well...
Oracle should push their changes to the repository... Could help to prevent duplicate work.

grooverdan:
@eworm-de feel your pain. it would be nice.

eworm-de:

PR #72, opened 2016-05-11
5.7 bug 81399 test replace number round+gis precision differences

grooverdan:
While testing gis functions on ppc64le I discovered that the double precision frequently caused 1.9999 -> 2 discrepancies that weren't corrected by the regex.

I've added the replace_numberic_round mysqltest function that round decimal output to the prescribed precision. Excessive 0s at the end of output are removed.

The inbuilt rounding of printf has caused rest results to be rounded (sign always assumed to be positive). The rounded results here are as a result of running on an x86 machine.

There are still some test failures on ppc64le after these commits due to slightly different geometries returned. I have some been given some gis advice as to which of these are incorrect and will solve the cause of these in a future pull request.


grooverdan:

PR #74, opened 2016-05-17
innodb buf_resize_thread - release mutexes when no change

grooverdan:
The Innodb buf_resize_thread fails to release the mutexes when the old and new innodb buffer pool sizes are the same.

While the innodb_buffer_pool_size_validate does actually do the same check, its theoretically possible to resize the pool twice and trigger this condition if both changes occur before innodb_buffer_pool_size_update is called the first time.


grooverdan:

grooverdan:
Its invalid - the mutexes are released already.

PR #76, opened 2016-05-24
bug #81399: Use numberic rounding in gis test cases (Part 2)

grooverdan:
--replace_numberic_round used for the following test cases
- gis.geometry_class_attri_prop
- gis.spatial_analysis_functions_distance
- gis.spatial_analysis_functions_centroid

This rounds gis results to 10 decimal places to avoid cross platform precision errors.

This depends on the replace_numberic_round implementation attached to the current #81399 bug.


grooverdan:

PR #78, opened 2016-05-27
Update split.c

ghost:
Buffer overflows can be triggered that when an attacker wrote payload to argv[1] value larger than 512 byte:
    if (argc > 4)
        for (n = atoi(argv[3]); n > 0; n--) {
            (void) strcpy(buf, argv[1]); //<==== buffer overflow
        }
or 
else if (argc > 3)
        for (n = atoi(argv[3]); n > 0; n--) {
            (void) strcpy(buf, argv[1]); //<==== buffer overflow
            (void) split(buf, fields, MNF, argv[2]);
        }

### DIFF
--- a/regex/split.c
+++ b/regex/split.c
@@ -1,5 +1,6 @@
 #include <stdio.h>
 #include <string.h>
+#define BUFFSIZE 512
 
 /*
  - split - divide a string into fields, like awk split()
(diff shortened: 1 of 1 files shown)


drmint80:
I have submitted OCA.

PR #80, opened 2016-06-08
Bug #80818: default log-tc-size too small on POWER - 3 * ( 64K page size) minimium needed

akopytov:
This is a slightly modified version of the patch contributed by Daniel
Black, adding support for AArch64 in addition to POWER when choosing the
page size for TC log.

### DIFF
--- a/mysql-test/include/mysqld--help.inc
+++ b/mysql-test/include/mysqld--help.inc
@@ -56,6 +56,8 @@ perl;
     s/\b196608\b/262144/;
     # Replacing port number with string
     s/^port \d+/port ####/;
+    # log-tc-size varies to 64K*3 on some platforms
+    s/^log-tc-size 262144/log-tc-size 24576/;
     foreach $var (@env) { s/$ENV{$var}/$var/ }
     next if /use --skip-(use-)?symbolic-links to disable/; # for valgrind, again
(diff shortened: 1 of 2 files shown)


akopytov:

setharnold:
I'm concerned that the defined value for two platforms is odd while the defined value for all the other platforms is a power of two. Is this a valid concern?

Thanks

tyhicks:
It looks like @akopytov modified the value in the original patch provided by @grooverdan from 65536 to 65535 but, from what I can tell, there's no public indication of why the modification was made. It would be great to know if this was intentional. Thanks!

akopytov:
@setharnold @tyhicks 65535 wasn't intentional. I don't remember whether that was a typo, or I just picked an early version of the patch by @grooverdan. Anyway, I noticed and fixed it later in my MySQL branch.

grooverdan:
Definitely typo - https://bugs.mysql.com/file.php?id=23825&bug_id=80818 I actually did 65536 and earlier I had 64*1024

Just yesterday I did ask Ubuntu (https://bugs.launchpad.net/ubuntu/+source/mysql-5.7/+bug/1706291) to carry the corrected patch of @akopytov or the backport in #160

PR #83, opened 2016-06-21
fix for innodb_numa_interleave missing

ghost:
Latest MySQL 5.6 misses the innodb_numa_interleave server variable.
https://bugs.mysql.com/bug.php?id=78953

### DIFF
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -663,5 +663,6 @@
 
 #cmakedefine CPU_LEVEL1_DCACHE_LINESIZE @CPU_LEVEL1_DCACHE_LINESIZE@
 #cmakedefine HAVE_LIBNUMA 1
+#cmakedefine WITH_NUMA 1
 
 #endif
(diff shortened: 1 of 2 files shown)


grooverdan:
so this is a copy of the patch in http://bugs.mysql.com/bug.php?id=80288

ghost:
Yes, that is correct. I verified the same fix is needed for the 5.6 branch and was hoping to get it fixed here as well.

ghost:

mysql-admin:
Hi Igroenewold,
Sorry but as noted above and through reading of bug http://bugs.mysql.com/bug.php?id=80288 this patch is not an original work but a copy of code written by a 3rd party person. As such it cannot be offered by you as your contribution under the terms of the OCA. As such I am closing this pull request

Thanks
==Omer

laurynas-biveinis:
@mysql-admin, would it work if I or my colleague who wrote the patch confirmed this as an OCA submission?

mysql-admin:
@laurynas-biveinis  your colleague (that wrote the patch) is currently not part of your 'group' OCA. He needs to be added there so we can associate the OCA with his bug system user. Then, he can submit it via a pull request similar to this one. 

Thanks
==Omer

PR #84, opened 2016-06-28
bug #82039: Add valgrind tool to MTR

grooverdan:
Using other valgrind tools in MTR is a little difficult as the MTR ties the tool to memcheck or calllgrind.

This provides a valgrind-tool option for easy selection.

MTR_TEST_NAME is also exported for recording results per test run (using a custom --valgrind-path= wrapper script).


grooverdan:

PR #87, opened 2016-07-30
SQLClient + SQL shim plugins

greenlion:
This merge incorporates the development branch 'internal_client'
- the internal SQL interface is introduced 
- SQL shim plugins are introduced


greenlion:
oops.

PR #88, opened 2016-08-03
bug 82427: Check filesize read in sql authentication plugin (5.7)

grooverdan:
Avoids compile error in gcc-5.4.0 (the default in Ubuntu-16.04)

sql/auth/sql_authentication.cc: In member function 'bool Rsa_authentication_keys::read_key_file(RSA**, bool, char**)':
sql/auth/sql_authentication.cc:220:60: error: ignoring return value of 'size_t fread(void_, size_t, size_t, FILE_)', declared with attribute warn_unused_result [-Werror=unused-result](void) fread(*key_text_buffer, filesize, 1, key_file);


grooverdan:

PR #89, opened 2016-08-03
bug 82428: Check filesize read in rsa authentication plugin (5.6)

grooverdan:
Avoids compile error in gcc-5.4.0 (default in ubuntu-16.04)

sql_acl.cc: In member function 'bool Rsa_authentication_keys::read_key_file(RSA**, bool, char**)':
sql_acl.cc:11688:62: error: ignoring return value of 'size_t fread(void_, size_t, size_t, FILE_)', declared with attribute warn_unused_result [-Werror=unused-result](void) fread(*key_text_buffer, filesize, 1, key_file);

### DIFF
--- a/sql/sql_acl.cc
+++ b/sql/sql_acl.cc
@@ -11659,12 +11659,19 @@ private:
       /* For public key, read key file content into a char buffer. */
       if (!is_priv_key)
       {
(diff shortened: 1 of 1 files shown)


grooverdan:

PR #91, opened 2016-08-23
Fix compile w/ G++ 6.1.1 on Fedora 24

dveeden:
storage/innobase/include/ut0dbg.h:52:20: error: nonnull argument ‘offsets’ compared to NULL [-Werror=nonnull-compare]

storage/innobase/fts/fts0fts.cc:6822:17: error: C++11 requires a space between string literal and macro [-Werror=c++11-compat]

sql/sql_update.cc:813:7: error: this ‘if’ clause does not guard... [-Werror=misleading-indentation]

A similar issue exists in boost:
boost_1_59_0/boost/numeric/ublas/matrix_expression.hpp:2224:17: error: this ‘if’ clause does not guard... [-Werror=misleading-indentation]

### DIFF
--- a/cmake/maintainer.cmake
+++ b/cmake/maintainer.cmake
@@ -22,7 +22,7 @@ SET(MY_C_WARNING_FLAGS
 
 # Common warning flags for G++ and Clang++
 SET(MY_CXX_WARNING_FLAGS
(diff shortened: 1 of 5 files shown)


dveeden:
This is with `-DCMAKE_BUILD_TYPE=Debug`

dveeden:

PR #92, opened 2016-08-28
Bug #82760: Do not use -fno-expensive-optimizations just to avoid fused madd instructions

akopytov:
-fno-expensive-optimizations is used on some architectures with the only
purpose of avoiding fused multiple-add instructions usage by the
compiler. That's too restrictive. Use -ffp-contract=off instead.

This also removes some IA64-specific --replace_result directives that
are unnecessary with this change and reasonably recent GCC
versions (it's just that neither -mno-fused-madd nor -ffp-contract=off
were supported by GCC on IA64 at the time those directives were added).


akopytov:

PR #94, opened 2016-09-03
Don't dump mysql.gtid_executed

dveeden:
https://bugs.mysql.com/bug.php?id=82848

Otherwise this will overwrite the value which is set with
SET @@GLOBAL.GTID_PURGED=...
And this could cause a server to lose the GTID position after a restart


dveeden:

dveeden:

PR #95, opened 2016-09-06
Add TLS SNI support to the client.

FGasper:
The shipped yassl library doesn’t seem to include SNI, but OpenSSL makes it easy to have this now.

I don’t know of any MySQL-compatible servers that implement SNI, but having it on the client now will make it that much more useful when a server implements it. It’s basically “free” for now, so there’s little reason not to have it.

I have a Perl test server script that demonstrates that this works, if that’s of use.

### DIFF
--- a/include/violite.h
+++ b/include/violite.h
@@ -208,7 +208,7 @@ struct st_VioSSLFd
 };
 
 int sslaccept(struct st_VioSSLFd*, Vio *, long timeout, unsigned long *errptr);
-int sslconnect(struct st_VioSSLFd*, Vio *, long timeout, unsigned long *errptr);
+int sslconnect(struct st_VioSSLFd*, Vio *, long timeout, unsigned long *errptr, const char *host);
 
 struct st_VioSSLFd
 *new_VioSSLConnectorFd(const char *key_file, const char *cert_file,
--- a/sql-common/client.c
+++ b/sql-common/client.c
@@ -3411,7 +3411,7 @@ Establishes SSL if requested and supported.
 @retval 1       failure
 */
 static int
-cli_establish_ssl(MYSQL *mysql)
+cli_establish_ssl(MYSQL *mysql, const char *host)
 {
 #ifdef HAVE_OPENSSL
   NET *net= &mysql->net;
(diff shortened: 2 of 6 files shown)


FGasper:

-FG

jawabuu:
I would like to follow up on this and get some feedback on the status.
https://bugs.mysql.com/bug.php?id=84849 and https://bugs.mysql.com/bug.php?id=82872 is pending for over 3 years now.
In my opinion this enhancement would provide huge enhancements in terms of routing, multi-tenancy, access-control and security by allowing user to manage access to their mysql instances through a single/unified entrypoint.
Traefik and HAProxy are especially equipped to handle this.

jawabuu:
> I would like to follow up on this and get some feedback on the status.
> https://bugs.mysql.com/bug.php?id=84849 and https://bugs.mysql.com/bug.php?id=82872 is pending for over 3 years now.
> In my opinion this enhancement would provide huge enhancements in terms of routing, multi-tenancy, access-control and security by allowing user to manage access to their mysql instances through a single/unified entrypoint.
> Traefik and HAProxy are especially equipped to handle this.

@FGasper any update on this?

FGasper:
@jawabuu I submitted this PR with the idea that, _when_ SNI support was added to MySQL Server, prior clients would already be compatible. So, for example, if this had merged in time for, say, MySQL 7, had MySQL 8 added SNI, then any libmysqlclient apps—e.g., PHP—that use client version 7 would automatically work with server version 8’s SNI.

The MySQL team saw things differently and had no interest in such forward compatibility. Perhaps they envision no need for adding SNI to MySQL Server; they would, after all, have to introduce some mechanism to fetch a certificate based on the SNI string, and it can be hard to implement that such that it suits everyone’s needs. Moreover, DB access over TCP is a much more esoteric thing than, say, HTTP or mail; a DB programmer can understand perfectly well why `brians-golf.com`’s DB is `server3.whizzo.bobs-hosting.com`. So perhaps there just isn’t enough need to justify the server-side development effort.

There’s also something to be said for the fact that the unencrypted SNI string is something of a security problem. Back when SNI support was new no one cared because it was a big improvement over status quo. But in our post-Let’s-Encrypt world, where TLS is pretty well ubiquitous, if an server doesn’t use the SNI string anyway, it’d be better for the client to withhold it. We might as well _not_ disclose the specific server name the client intends to hit if all things else are equal.

PR #97, opened 2016-09-07
Bug #82889: DebugPrintDeathTest.PrintEval test failure.

akopytov:
Commit 5c2b61d has brought two unit tests (DebugPrintTest and
DebugPrintDeathTest) that use "division by zero" exception to check
the evaluation of args of DBUG_PRINT macro.

C99 says in 6.5.5p5: "... if the value of the second operand is zero,
the behavior is undefined".

Also, according to the ARMv8 Architecture Reference Manual if the divisor
is zero, the result of operation is zero and no exception is generated.
Consequently, the DebugPrintDeathTest.PrintEval unit test fails.

This patch avoids using "division by zero" exception.

### DIFF
--- a/unittest/gunit/dbug-t.cc
+++ b/unittest/gunit/dbug-t.cc
@@ -80,23 +80,19 @@ TEST(DebugPrintTest, PrintEval)
   int y= 0;
 
   // This DBUG_PRINT args should never be evaluated.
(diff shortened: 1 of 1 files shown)


akopytov:

mysql-admin:
Hi Alexey.
This pull request includes commits for another person, it cannot be accepted under your OCA
==Omer

PR #98, opened 2016-09-07
Bug #82890: RDTimeStampCounter.TestCycle test failure.

akopytov:
RDTimeStampCounter.TestCycle unit test expects that the cycle
timer counter will overflow. Aarch64 cycle timer implementation
uses CNTVCT_EL0 register which has a width of 64 bits, so do not
expect timer overflow during the test.


akopytov:

mysql-admin:
Hi Alexey.
This pull request includes commits for another person, it cannot be accepted under your OCA
==Omer

PR #100, opened 2016-09-08
Bug #82890: RDTimeStampCounter.TestCycle test failure.

akopytov:
RDTimeStampCounter.TestCycle unit test expects that the cycle timer
counter will overflow. The AArch64 cycle timer implementation uses
a 64-bit register, so it is unreasonable to expect overflow, whereas the
condition in the unittest expects at least 1 overflow for some reasons.

Fixed by changing the unittest to not expect an overflow which was
likely the original intent.


akopytov:

akopytov:

PR #102, opened 2016-09-19
Bug#20671150 BROKEN BUILD DUE MISSING CHECK OF RETURN VALUE OF FREAD()

grooverdan:
Check return value from fread(), return error if it fails.

Ported to 5.6 by Daniel Black

Signed-off-by: Daniel Black daniel.black@au.ibm.com

### DIFF
--- a/sql/sql_acl.cc
+++ b/sql/sql_acl.cc
@@ -11678,17 +11678,26 @@ private:
       }
 
       /* For public key, read key file content into a char buffer. */
+      bool read_error= false;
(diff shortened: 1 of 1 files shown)


grooverdan:

mysql-admin:
Hi grooverdan,

Thanks for the contribution. The fix mentioned in bug http://bugs.mysql.com/bug.php?id=82427 will be back-ported to 5.6 in an upcoming release 

==Omer

PR #103, opened 2016-09-19
Bug#20671150 BROKEN BUILD DUE MISSING CHECK OF RETURN VALUE OF FREAD()

grooverdan:
Check return value from fread(), return error if it fails.

Not sure why this didn't end up in the latest 5.7 release. Guess it wasn't fixed in 5.7 branch.


grooverdan:

mysql-admin:
Hi grooverdan,
Thanks for the contribution but this bug has been fixed and closed already including a fix in 5.7.
Please see http://bugs.mysql.com/bug.php?id=82427
==Omer

PR #104, opened 2016-09-19
Innodb: innodb_numa_interleave=1, use numa numa_get_mems_allowed

grooverdan:
Using numa_all_nodes_ptr was excessively optimistic. Due to
constraints in systemd, containers or otherwise MySQL could of been
limit to a smaller set of cpus. Use the numa_get_mems_allowed
library function to see what we can interleave between before doing
so.

Signed-off-by: Daniel Black daniel.black@au.ibm.com


grooverdan:

PR #106, opened 2016-10-18
Add new scheduling algorithm for reducing tail latencies (VATS 1.0)

sensssz:
This branch introduces a new scheduling algorithm (Variance-Aware-Transaction-Scheduling, VATS) for the record lock manager of InnoDB based on MySQL 5.7. Instead of using First-Come-First-Served (FCFS), the newly introduced algorithm chooses between FCFS and Eldest-Transaction-First (ETF), which prefers older transactions over new ones. A configuration parameter (innodb_lock_schedule_algorithm) is introduced for users to choose between VATS and FCFS (the default one). Another parameter (innodb_vats_wait_lock_pct) is introduced to specify the threshold of the percentage of wait locks of choosing ETF over FCFS. We've extensively tested this algorithm in many workloads. The algorithm is very simple, and the changes are very local, but it significantly improves performance (in terms of average latency and throughput) and predictability (in terms of reduction of tail and quantile latencies) For more details, please refer to this paper http://arxiv.org/abs/1602.01871

### DIFF
--- a/sql/rpl_rli_pdb.cc
+++ b/sql/rpl_rli_pdb.cc
@@ -94,6 +94,7 @@ bool handle_slave_worker_stop(Slave_worker *worker,
     worker->running_status= Slave_worker::STOP_ACCEPTED;
     mysql_cond_signal(&worker->jobs_cond);
     mysql_mutex_unlock(&rli->exit_count_lock);
+    is_slave_replication = false;
     return(true);
   }
   else if (rli->exit_counter == rli->slave_parallel_workers)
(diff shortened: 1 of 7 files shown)


grooverdan:
An is_slave_replication boolean is the wrong way to do this. There is nothing architecturally preventing a slave from receiving updates directly from a client rather than the master.

Have you considered that 5.7 has multisource replication?  https://dev.mysql.com/doc/refman/5.7/en/replication-multi-source.html

See the thd_is_replication_slave_thread(trx->mysql_thd) function as this probably does what you want though I haven't quite grasped why replication transactions are treated differently.

Thanks for doing a 5.7 patch.

sensssz:
@grooverdan Hi, Daniel. Thanks for your suggestion. I've replaced is_slave_replication with thd_is_replication_slave_thread. I disable VATS for replication because replication transactions are required to commit "in order", so changing the order of the transactions may not be as effective and is not what VATS is designed for. We think it safer to disable it during replication now.

sensssz:

PR #107, opened 2016-10-20
Bug #83457: Incorrect ISA detection code in OEL RPM spec

akopytov:
packaging/rpm-oel/mysql_config.sh:
- fix code around "rpm --eval %__isa_bits" to work as intended
- add "aarch64" to the list of 64-bit architectures in the fallback code

packaging/rpm-fedora/mysql_config.sh:
- make it identical to the same script in packaging/rpm-oel/


akopytov:

PR #108, opened 2016-11-04
Bug #79151 - error -1 from storage engine while creating table with non-existing datadir

elshadaghazade:
According to this bug report http://bugs.mysql.com/bug.php?id=79151
I added three lines of codes to send a client an occured system error code, but not unclear -1 error.


elshadaghazade:

PR #110, opened 2016-11-11
Bug #71879 - Wrong binlog size reported after moving binlogs

elshadaghazade:
Problem was that after moving the bin log files to a new folder, real path of these files was wrong, for this reason program couldn't open and calculate its' size. I corrected files with wrong path, prepended full path of new binlog directory to these files.


elshadaghazade:

elshadaghazade:
Please accept only changes on sql/rpl_master.cc file. Updates on the file fil0fil.cc are from previous bug fix which I sent already.

PR #111, opened 2016-11-12
Bug #83741 - InnoDB: Failing assertion: lock->magic_n == 22643

elshadaghazade:
I did not find any explanation or documentation about ib_create_table_fail_at_create_index debug parameter. 

DBUG_EXECUTE_IF(
  "ib_create_table_fail_at_create_index",
   len = DICT_MAX_FIELD_LEN_BY_FORMAT(table) + 1;
);

It seems these debug condition was used for testing some errors during development, but was forgotten to delete later.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -21,6 +21,8 @@
 #include <functional>
 #include <list>
 #include <set>
+#include <unistd.h> // this library will help us to detect file system status
+#include <string.h>
 #include <string>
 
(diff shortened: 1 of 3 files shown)


akopytov:
ib_create_table_fail_at_create_index is used by mysql-test/suite/innodb/t/innodb_wl6469_debug.test for error injection.

ShahriyarR:
@akopytov so this must be only affected with Temp tables?

PR #112, opened 2016-11-26
Report timing with more precision

dveeden:
```
mysql> select sleep(0.123);
+--------------+
| sleep(0.123) |
+--------------+
|            0 |
+--------------+
1 row in set (0.123 sec)
```

### DIFF
--- a/client/mysql.cc
+++ b/client/mysql.cc
@@ -5537,8 +5537,9 @@ static ulong start_timer(void)
 #if defined(_WIN32)
   return clock();
 #else
-  struct tms tms_tmp;
-  return times(&tms_tmp);
+  struct timespec start;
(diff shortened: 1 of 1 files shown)


dveeden:
The other option is to change the return type of start_timer from ulong to double.

dveeden:

grooverdan:
since the client is linked to ../libmysql/libmysqlclient.a which contains mysys, using my_timer_init/my_timer_milliseconds/my_timer_microseconds/ my_timer_nanoseconds from mysys/my_rdtsc.c might be the neatest cross platform solution.

dveeden:
@grooverdan I agree. Closing this PR in favour of #113

PR #113, opened 2016-11-28
More exact timing for mysql client based on my_timer_microseconds

dveeden:
Based on suggestion from @grooverdan on https://github.com/mysql/mysql-server/pull/112


dveeden:

grooverdan:
nice. The 10000 division looks odd but I believe you

PR #114, opened 2016-12-01
change (isinf || isnan) to !isfinite

grooverdan:
my_isinf(square) || my_isnan(square) -> !my_isfinite(square)

my_isfinite is a single function call and may by optimised more than two function calls. The logic is equivalent as finite, infinite and nan are the only states.


grooverdan:

PR #115, opened 2016-12-15
New version of the VATS algorithm (Largest Dependency Set First)

sensssz:
This pull request uses a new version of the VATS algorithm under version 5.7.


grooverdan:
what is different about the new implementation?

sensssz:
Hi, Daniel. In this new implementation we look at the sub-tree of each lock instead of just looking at the arrival time in the system. This gives us a more accurate yet efficient enough way of deciding which transaction to schedule first.

sensssz:

grooverdan:
@sensssz , great, keep up the good work! :+1:

PR #116, opened 2016-12-21
mtr: mem is pure flag / add /run/shm

grooverdan:
Since 716621db3f3055781e24f561325cec6eac181717, opt_mem is
a pure flag. However it is assigned to the value of $MTR_MEM.

This leads to conflicts when MTR_MEM=/xxx/yy ./mtr --mem is
invoked. Here the --mem option overrided opt_mem leaving the
default path to be chosen.

This change makes when MTR_MEM set, opt_mem, the flag, is also
set. Both the environment and flag can no be set without conflicting.

/run/shm is added to the beginning of the path as its now used
in a few distros and is conceptually better than /dev for its purpose.

### DIFF
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -219,7 +219,7 @@ sub using_extern { return (keys %opts_extern > 0);};
 
 our $opt_fast= 0;
 our $opt_force;
-our $opt_mem= $ENV{'MTR_MEM'};
+our $opt_mem= $ENV{'MTR_MEM'}  ? 1 : 0;
 our $opt_clean_vardir= $ENV{'MTR_CLEAN_VARDIR'};
 
 our $opt_gcov;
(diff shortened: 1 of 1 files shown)


grooverdan:

bjornmu:
I'm afraid it's a bit more complicated than just adding /run/shm to the list of alternatives. On Ubuntu 14.04 they have changed the ramdisk to /run/shm but have kept /dev/shm as a softlink. This was causing some tests to fail because they would compare a path as known by mtr with /dev/shm with one known by the server. But the path reported by the server had been canonicalized to /run/shm and this did not match. With your patch this would not be a problem as it will use /run/shm.

However, sometime between Ubuntu 14.04 and 16.04 they switched back to using /dev/shm and now /run/shm is a softlink instead. Since /run/shm exists and is first in your list mtr will use that and now some tests will fail because the server uses /dev/shm.

What mtr really needs to do is pick the first path in the list which is *not* a softlink. So adding /run/shm to the list is good but you'll also need this additional check.

PR #118, opened 2017-01-01
Print binary data as hex in the mysql client

dveeden:
Print binary data as hex literals. 
The result is that this doesn't break my terminal if I select a binary column and this allows me
to copy-paste the output to the were clause of my next query.

```
mysql> select * from t1;
+----+------------------------------------+
| id | ip                                 |
+----+------------------------------------+
|  1 | 0x00000000000000000000000000000001 |
|  2 | 0x7F000001                         |
|  3 | 0x08080808                         |
|  4 | 0x08080404                         |
+----+------------------------------------+
4 rows in set (0.00 sec)

mysql> show create table t1\G
*************************** 1. row ***************************
       Table: t1
Create Table: CREATE TABLE `t1` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `ip` varbinary(16) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=5 DEFAULT CHARSET=latin1
1 row in set (0.00 sec)

mysql> select ip,INET6_NTOA(ip) FROM t1 WHERE ip=0x08080404;
+------------+----------------+
| ip         | INET6_NTOA(ip) |
+------------+----------------+
| 0x08080404 | 8.8.4.4        |
+------------+----------------+
1 row in set (0.00 sec)
```

### DIFF
--- a/client/mysql.cc
+++ b/client/mysql.cc
@@ -4170,6 +4170,14 @@ print_table_data_xml(MYSQL_RES *result)
 }
 
 
+static void print_blob_as_hex(FILE *output_file, const char *str, ulong len)
+{
+    const char *ptr= str, *end= ptr + len;
+    for (; ptr < end ; ptr++)
+      fprintf(output_file, "%02X", *((uchar *)ptr));
+}
+
+
 static void
 print_table_data_vertically(MYSQL_RES *result)
 {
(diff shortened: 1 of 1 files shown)


dveeden:

PR #120, opened 2017-02-16
C++ conformance fixes (MSVC /permissive-)

Rastaban:
We (the Microsoft C++ team) use the MySQL project as part of our "Real world code" tests.
I noticed a few issues in windows specific code when building MySQL with the MSVC compiler
in its conformance mode (/permissive-).  For more information on /permissive- see our blog
https://blogs.msdn.microsoft.com/vcblog/2016/11/16/permissive-switch/.

These changes are to address 3 different types of issues:

1) Use of qualified names in member declarations

    struct A {
        void A::f() { } // error C4596: illegal qualified name in member declaration
                        // remove redundant 'A::' to fix
    };

2) String literal type conversion

    char * str = "string"; // String literal loses 'const'
	                       // either change to 'const char*' or cast literal to 'char*'

I cast the value for shared_memory_base_name because adding 'const' to something like that
can have a big domino effect and I'm not sure I have the resourses to verify a change like that on
other systems.  You may want to make it a low priority 'todo' item for the future.  For now
it makes the current compiler behavior explicit.

3) Do not allow skipping an initializer with a goto

    void func(bool b)
    {
        if(b) goto END;
  
        int value = 0; //error C2362: initialization of 'value' is skipped by 'goto END'
              int array[10]; //Okay, not initialized.
  
        //... value used here
  
    END:
        return;
    }

One potential fix is to limit the scope of the initialized variable so it ends before the label.
See our blog for more potential fixes.

### DIFF
--- a/client/mysql_secure_installation.cc
+++ b/client/mysql_secure_installation.cc
@@ -35,7 +35,7 @@ static my_bool g_expire_password_on_exit= FALSE;
 static my_bool opt_use_default= FALSE;
 
 #if defined (_WIN32) && !defined (EMBEDDED_LIBRARY)
-static char *shared_memory_base_name= default_shared_memory_base_name;
+static char *shared_memory_base_name= (char*)default_shared_memory_base_name;
 #endif
 
 #include "sslopt-vars.h"
--- a/libmysql/authentication_win/handshake.h
+++ b/libmysql/authentication_win/handshake.h
@@ -100,7 +100,7 @@ public:
   Handshake(const char *ssp, side_t side);
   virtual ~Handshake();
 
-  int Handshake::packet_processing_loop();
+  int packet_processing_loop();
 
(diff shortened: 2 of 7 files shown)


client/mysql_secure_installation.cc
```diff
@@ -35,7 +35,7 @@ static my_bool g_expire_password_on_exit= FALSE;
 static my_bool opt_use_default= FALSE;
 
 #if defined (_WIN32) && !defined (EMBEDDED_LIBRARY)
-static char *shared_memory_base_name= default_shared_memory_base_name;
+static char *shared_memory_base_name= (char*)default_shared_memory_base_name;
```
foobar, line 38: Should this be `const_cast`?

Rastaban: Yes

PR #122, opened 2017-03-01
Bug #81814: InnoDB adaptive hash index uses a bad partitioning algorithm for the real world

akopytov:
This patch replaces hash-based AHI partitioning with index ID based
one.

Index ID based partitioning is better than the current bad hash
function. It is also not any worse than any 'good' hash function for
real world workloads.

Adding space IDs would only make sense if index IDs were not guaranteed
to be unique. They are unique in 5.7 and the current public release of
8.0.


akopytov:

PR #123, opened 2017-03-01

akopytov:
Use PRNG-based slot indexes for InnoDB row counters on AArch64.

Using RDTSC- (aka counter-based) indexes result in higher contention on
many-core AArch64 machines due to low timer resolution (100 MHz). Using
sched_getcpu()-based indexers leads to lower single-threaded
performance, because sched_getcpu() is expensive on AArch64 (it is not
implemented via VDSO like on x86).

This also requires a better quality PRNG and properly seeding it for
each thread. These will be addressed in separate patches.


akopytov:

PR #125, opened 2017-03-05
Contribution for bug #83648

robgolebiowski:
The bug was caused by three problems:
1) query->intersection was not freed in case of error caused by
exceeding innodb_ft_result_cache_limit.
2) errors from init_ftfuncs were not propagated - this was fixed
in 5.7 by bug fix 21140111. This was ported into 5.6
3) bug fix 21140111 was causing assertion failure when
innodb_ft_result_cache was exceeded in DELETE command. This was
also fixed.


robgolebiowski:

grooverdan:
it appears that bug 83648 is access controlled making following updates cumbersome.

PR #127, opened 2017-03-15

akopytov:
This is a backport of commit 32b184c from 8.0 (which has been reverted from lower versions for unknown reasons) + a change to seed thread-local PRNG with OS thread ID. This is required to optimize InnoDB statistic counters on many-core ARM64 machines.


akopytov:

PR #129, opened 2017-03-17

akopytov:
This patch adds a new set of os_atomic_*() macros:

- os_atomic_*_nr() to atomically add/subtract a value to/from a variable
  without returning its original value;

- os_nonatomic_*_nr() to do the same non-atomically.

This semantics allows hardware-specific optimizations on some
architectures. For example, the STADD instruction available on AArch64
CPUs with LSE (Large System Extensions) performs an atomic
addition/subtraction while discarding the original value (i.e. not
loading it into a register). Which not only results in better
scalability as compared to the standard LL/SC synchronization mechanism
provided by AArch64 without LSE, but is also faster than the regular
non-atomic add/subtract instructions on AArch64, since the standard
load/modify/store sequence is replaced with a single instruction,
similar to the code generated for x86(_64).

This kind of optimizations should theoretically be performed by the
compiler, but compiler support is lagging behind even in the most recent
GCC versions.

This patch maps the new family of macros to the optimized LSE-based
implementations on AArch64. It also changes InnoDB fuzzy counters to use
the "non-atomic no-return add/subtract" semantics. No changes in
behavior is introduced for other architectures, i.e. regular
add/subtract code is emitted.

The scope of this optimization is not limited to InnoDB fuzzy counters.
There are many other counters that do not require the original value
when updated. But those will be addressed separately based on the
primitives introduced by this patch.

### DIFF
--- a/configure.cmake
+++ b/configure.cmake
@@ -924,6 +924,11 @@ IF(HAVE_LIBNUMA AND NOT WITH_NUMA)
    MESSAGE(STATUS "Disabling NUMA on user's request")
 ENDIF()
 
+OPTION(WITH_LSE "Enable Large System Extensions for AArch64" OFF)
+IF(WITH_LSE)
+  ADD_DEFINITIONS(-DHAVE_ARM64_LSE_ATOMICS)
+ENDIF()
+
 # needed for libevent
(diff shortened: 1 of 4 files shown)


akopytov:

PR #131, opened 2017-03-29
Bug #84415 slave don't report Seconds_Behind_Master when

robgolebiowski:
running slave_parallel_workers > 0

The fix: When a task from GAQ finishes executing, last_master_timestamp
gets assigned timestamp from next task in GAQ. In case GAQ is empty
last_master_timestamp gets reset to 0. Because last_master_timestamp
is only re-assigned when task leaves GAQ it must be also assigned
when first event (from parallel execution) arrives.


robgolebiowski:

robgolebiowski:
This PR introduces valgrind failure, will submit a correct patch shortly.

PR #143, opened 2017-05-15
Bug#86321: Test binlog.binlog_gtid_state_update_deadlock is

robgolebiowski:
failing on 5.7 trunk

According to this test sync point after_log_file_name_initialized
should be initialized on changing GLOBAL.GTID_MODE. However, debug sync
after_log_file_name_initialized  was never hit on changing GTID_MODE.
The purpose of this debug sync point was to hold execution after
gtid_mode_lock was acquired, but before global_sid_lock was locked.
The fix was to add new debug sync
gtid_mode_update_gtid_mode_lock_wrlock_taken_will_take_global_sid_lock
which fulfills this requirements. Without the fix the test just
timeouts on debug sync point after_log_file_name_initialized – that is why
–include/big_test.inc was added to this test, which now
has been removed.


robgolebiowski:

mysql-admin:
Hi,
Thanks for submitting the contribution. Please not that https://bugs.mysql.com/bug.php?id=86321 was determined to be "can't repeat", so a contribution for it cannot be accepted

Thanks
==Omer

mysql-admin:
Closing request as bug could not be reproduced

PR #146, opened 2017-06-01
Bug #85588: Regular expressions work incorrectly for strings having 00 byte

Olernov:


Olernov:
https://bugs.mysql.com/bug.php?id=85588
When using rlike operator to detect characters with hex values between 80 and FF, it does not work when converting a binary string that starts with hexadecimal '00' using unhex function.

Regex library uses strlen function so it does not work correctly for binary strings having 00 byte. Issue is solved by passing string length to avoid using strlen.

Olernov:

PR #147, opened 2017-06-02
Bug #85588: When using rlike operator to detect characters with hex v…

Olernov:
…alues between 80 and FF, it does not work when converting a binary string that starts with hexadecimal '00' using unhex function.


Olernov:

PR #149, opened 2017-06-27
Fix Bug#86835 session_connect_attrs.processlist_id should be bigint

dveeden:
Untested yet.


dveeden:

PR #152, opened 2017-06-29
Bug #86865: InnoDB does unnecessary work when extending a tablespace

akopytov:
Do not call fil_write_zeros() in fil_space_extend() if posix_fallocate()
is available and was successful.


akopytov:

PR #153, opened 2017-07-02
Bug #86898: Suboptimal String::append() usage in JSON processing

akopytov:
Amortize String::append() costs by collecting sequences of characters
that don't need escaping in the input string, and appending them to the
output string with a single String::append() call.


akopytov:

PR #156, opened 2017-07-19
include my_dir.h

eworm-de:
Building client libraries only (-DWITHOUT_SERVER=ON) failed with:

error: unknown type name ‘MY_STAT’

and

implicit declaration of function ‘my_fstat’

Include my_dir.h to fix compilation.


eworm-de:

PR #158, opened 2017-07-20
mtr: use replace_numeric for more gis tests

grooverdan:
spatial_analysis_functions_buffer revert to original now that
replace_numeric can show result differences.


akopytov:
I'm not sure `replace_numeric_round` is the right solution to the problem. I had a lot of fun with GIS tests during my refactoring of floating point in the server and inclusion of dtoa.c. All of those problems were actually results of wrong compiler flags on particular platforms (IA64, Windows, etc.) In some rare cases no support for compiler flags existed at the time (e.g. GCC did not provide any way to disable fused multiply-add instructions on IA64) so I had to resort to tricks like `replace_result`.

I understand floating point implementation on POWER has some extensions. But I'm curious, what's the real reason for results difference in these cases? Is it something that can be fixed with compiler flags or CPU initialization similar to `setup_fpu()` in `mysqld.cc`?

grooverdan:
I was just of the assumption that float numbers had some aspect of precision that was architecture specific and expecting the same answer at the 10th+ decimal place wasn't portable.

The fesetround has a implementation for POWER.

This wouldn't be https://bugs.mysql.com/bug.php?id=35071 in a form would it?

grooverdan:
I'll check if the replace results of gis-precise.test can be removed and try to minimise this change.

akopytov:
One can expect the same answer on all IEEE 754 architectures provided that no extensions to the standard are used. Some architectures may use higher (e.g. 80-bit) precision to represent floating point numbers internally. That fact, combined with [fused CPU instructions](https://en.wikipedia.org/wiki/Multiply–accumulate_operation) may indeed result in floating point calculations being non-portable. Which is a more general problem than just regression test failures. Consider import/export or statement-based replication, for example. So IMHO cases like this should be handled on a different level than just silencing test differences, if at all possible.

akopytov:
> This wouldn't be https://bugs.mysql.com/bug.php?id=35071 in a form would it?

Thanks, that bug should be fixed as of 5.5, I added a comment.

akopytov:
@grooverdan this commit may also be of interest: https://github.com/mysql/mysql-server/commit/0a417e84

grooverdan:
I make the follow patch with test cases recorded on x86_64.

https://github.com/grooverdan/mysql-server/commit/3fd6a93d7719a8c9b9d52274278b093ab917812d

The test case failures for GIS tests on ppc64le are as follows:

https://gist.github.com/grooverdan/3ba0aa21bf65fa59e12c67ebccfd2f18

Both where compiled with Ubuntu 16.04's GNU GCC 5.4.0 and ppc64le picked up the -ffp-contract=off  flag. No -fexpensive-optimizations where used on either compile.

What is the delta on ARM64?

akopytov:
There's no delta on ARM64, and I have never seen any test failures resulting from floating point calculations. This is from unmodified 5.7.19 on an ARM64 machine: https://gist.github.com/akopytov/2a29a7455ab95f03416dd84255ff98ac

GCC 5.4.0, -fp-contract=off, -fexpensive-optimizations.

grooverdan:
Does it still pass identically to x86_64 if all the replace_result/regex/numeric are removed fromgis test https://github.com/grooverdan/mysql-server/commit/3fd6a93d7719a8c9b9d52274278b093ab917812d#diff-a3fd1e7fcbafa92212ba7f207fccc3faL1270

If it does then it is all fixed and I can submit a version of the above commit selfishly modified to allow the 8 test cases I see on Power to pass (and I'll allow for ARM64 differences too if the exist). Is there other cases to consider (older compilers/architectures?).

akopytov:
Sorry, I didn't notice your changes in https://github.com/grooverdan/mysql-server/commit/3fd6a93d7719a8c9b9d52274278b093ab917812d#diff-a3fd1e7fcbafa92212ba7f207fccc3faL1270 originally. With that diff, I see the following 2 test failures on ARM64: https://gist.github.com/akopytov/707c49fcd57306e80aa764b5abbb7022

Everything else in gis.* and main.* passes.

Which as I understand is identical to Power, rather than x86_64. Which is even more suspicious, because it likely means the root cause is not in the floating point implementation per se.

The problem goes back further than I thought. I didn't know they started using `--replace_regex` in 5.7 in GIS tests on all architectures to hide rounding errors on specific ones. Like here: https://github.com/mysql/mysql-server/commit/0e6c8de34814

The proper approach would be to investigate rounding differences on specific architectures and fix them when possible. Otherwise, introduce a more flexible mechanism that would only tolerate rounding differences on those specific architectures. The current approach is dangerous IMHO, but I don't have bandwidth to investigate it further.

grooverdan:
Thanks for the tests. Much appreciate your donated bandwidth and pointers to key aspects in the history of gis precision. FWIW I agree with you that a better analysis of architecture precisions differences needs to occur. Given the use of -ffp-contract=off exists in 5.7 (cmake/floating_point.cmake) seems to be after the rounding kludges introduced I'll remove as much as possible, but due to bandwidth will leave rounding differences in x86_64 to another day/person.

grooverdan:
closing in favour of #162

PR #159, opened 2017-07-24
merge (#2)

addstone:
* Bug#22671846: SENDING LARGE MESSAGES CAN TRIGGER NODE EXIT FROM GROUP

Second attempt.

Problem:

When large messages are sent, the XCOM thread becomes busy for a long
time sending it. As this happens, mainly as the number of member
nodes grows, the alive task can trigger a group view change and GCS
can trigger a node exit due to the 5 second timeout period.

Solution:

For large messages, the select will mark the socket as readable many
times for a single message, since TCP sends a large message as
multiple fragments anyway. Each time this happens, the
acceptor_learner task is activated and will call read_msg/read_bytes.
Until now we have updated only when a complete messages has been
received, which may take a long time. By updating the server timestamp
inside the loop in read_bytes/buffered_read_bytes, the timestamp will
be updated often enough that nodes are not marked as being
unavailable.

* Bug #25189192	ERRORS WHEN RESTARTING MYSQL AFTER RENAME TABLE.

PROBLEM

While renaming table innodb doesn't update the InnoDB Dictionary table
INNODB_SYS_DATAFILES incase there is change in database while doing
rename table. Hence on a restart the server log shows error that it
couldnt find table with old path before rename which has actually been
renamed. So the errors would only vanish if we update the system
tablespace

FIX

Update the innodb dictionary table with new path in the case there is
not a change in the table but the database holding the table as well.

Reviewed-by: Jimmy Yang<Jimmy.Yang@oracle.com>
RB: 15751

* Bug#25479538  ASSERT:SIZE == SPACE->SIZE DURING BUF_READ_AHEAD_RANDOM

The assert as added by the fix for BUG#25053705  is blocking the ntest. So removing
the assert for now until I come up with a fix for the bug.

Reviewed-by : Satya Bodapati <satya.bodapati@oracle.com>

* Bug#24609402 MYSQLBINLOG --RAW DOES NOT GET ALL LATEST EVENTS

Problem: When mysqlbinlog is used with --raw option, output file
         is not getting flushed till the process is completed.
         With --stop-never option (mysqlbinlog process never completes),
         output file will never contain any events.

Analysis & Fix: --read-from-remote-server is a mysqlbinlog tool option that
                allows the users to read the binlog from a MySQL server
                rather than reading a local file.
                The read binary log event will be processed and
                written to output files in text format. In this case,
                the output file is getting flushed after every event in
                copy_event_cache_to_file_and_reinit() function.

                By default, mysqlbinlog reads binary log files and writes
                events in text format. But the --raw option tells
                mysqlbinlog to write the events into output file directly
                in their original binary format. In this case, the output
                file is not getting flushed until mysqlbinlog downloads all
                the content. This will be more problematic if --stop-never
                option is used (which is used for "live" binlog backup) as
                the mysqlbinlog process will never stop, the output file
                will never get flushed.

Fix: Flush the output file after every event is written to it
     (just like how we flush when --raw is not specified).

* Bug#24609402 MYSQLBINLOG --RAW DOES NOT GET ALL LATEST EVENTS

Fixing post push pb2 failure

* Bug #25501659: HANDLE_FATAL_SIGNAL (SIG=11) IN GET_ADDON_FIELDS

ISSUE: Uninitialized Field member of Table structure was getting
       dereferenced in function get_addon_fields() resulting in
       server exiting abnormally.

FIX: Passed pointer to Table structure to get_addon_fields() to
     get read_set from it directly.

* Bug #25573565   TABLE REBUILD USES EXCESSIVE MEMORY

Problem:
=======
   Offsets allocates memory from row_heap even for deleted row
traversal during table rebuild.

Solution:
=========
  Empty the row_heap even for deleted record. So that
offsets don't allocate memory everytime.

Reviewed-by: Jimmy Yang <jimmy.yang@oracle.com>
RB: 15694

* Bug#25674926 VIEW CHANGE LOG EVENTS GENERATE NEW GTIDS

Problem: When there is replication (async) between
         two groups, group nodes are confused when it
         receives view_change_log_event that is generated
         on the other group. This event is making round
         trips between these two groups in case if
         the async replication is circular replication.

Analysis: GR plugin pipeline considers that
          View_change_log_events are only injected on
          the pipeline by the plugin itself when a view
          change happens. On that moment, the pipeline
          wrap the View_change_log_event on a GTID and
          queue it on group_replication_applier channel.
          View_change_log_events that go through
          group_replication_applier and group_replication_recovery
          channels are not injected back on the plugin applier,
          that would cause a infinite loop. But Other channels
          are injecting the GTID, BEGIN, VIEW, COMMIT into the
          plugin pipeline. Then pipeline wraps again the VIEW
          on a new GTID and queues it on group_replication_applier
          channel. This is creating a infinite loop between two
          groups.

Fix: When a GTID, BEGIN, VIEW, COMMIT is injected on the pipeline,
    this VIEW must not originate a new GTID2, BEGIN, VIEW, COMMIT on
    group_replication_applier channel.

* BUG#25365223 ERRORLOG UPDATED WITH NOTES INFORMATION WHEN A REF
FKEY ISN'T FOUND IN GRSETUP

Issue
====
The issue is that the info message that InnoDB prints when a table
is created with a reference which doesn't exist fills up the log as
it's printed for every insert when foreign_key_checks is disabled.

Fix
===
The fix is to display the message only if foreign_key_checks is
enabled.

Reviewed-by: Jimmy Yang <jimmy.yang@oracle.com>

* WL#10520: Backport WL#9837 to MySQL 5.7.19

This worklog is created for the backport of the WL#9837: "Group Replication:
Transaction savepoint support" to MySQL 5.7.19

This worklog goal is to improve Group Replication by adding support to
SAVEPOINT.

The SAVEPOINT statement sets a named transaction savepoint with a name of
identifier. If the current transaction has a savepoint with the same name, the
old savepoint is deleted and a new one is set.

The ROLLBACK TO identifier command reverts the state of the transaction back to
what was when executed the command SAVEPOINT identifier.

The RELEASE SAVEPOINT statement removes the named savepoint from the set of
savepoints of the current transaction. No commit or rollback occurs. It returns
an error if the savepoint does not exist.

* test in all.

* BUG#25719975 SHEBANG HARD CODED AS /USR/BIN/PERL IN SCRIPTS, BREAKS ON FREEBSD

Use cmake variable to adjust shebang to platform.

* Bug#25795538 SEGFAULT-T FAILING UNDER RECENT ADDRESSSANITIZER

Patch for 5.6 only.

Disable Segfault test for ASAN builds.

Fix mismatching new/delete for Fake_TABLE objects.

Backport patch for:
Bug#21255860: ASAN: MEMORY LEAK IN QUEUE UNIT TEST

* Bug#25349924 MTR: FIXES AND ENHANCEMENTS TO THE --XML-REPORT OPTION

Issues
------
The XML report generated by MTR when the --xml-report option is used has
the below problems :-
- It does not report any test which has failed in one of the retries
  but has passed on the final attempt as a failure i.e., it is
  reported like a successful test making usage of the option with
  --retry > 1 ambiguous
- It does not distinguish between disabled and skipped tests
- The xml reports are generated in the source directory 'mysql-test'
  even for out-of-source builds
- Testcase tags lack information about combinations.
- There can be multiple <testsuite tags> for a single suite scattered
  accross the report such that each tag has a bunch of tests belonging
  to a suite.
- The <failure> tag does not include any information regarding failed tests

Changes
-------
The patch does the following :-
- Reports tests failing on any retry attempt with the <failure> tag
- Adds statistics and fields for skipped tests and disabled tests separately
- Creates xml report in build directory if no absolute path is given for
  out-of-source builds
- Adds a field named variation in the <testcase> tag in case it is a
  combination run
- Aggregate results of all tests belonging to a suite within a single
  <testsuite> tag
- Generates and reports information regarding failures with a brief reason
  in an attribute named 'message' and embeds other details within the tag

Reviewed-by: Pavan Naik <pavan.naik@oracle.com>
RB: 15016

* BUG#18949364 I_BINLOG.BINLOG_MYSQLBINLOG_FILE_WRITE FAILS: COULD NOT
             CONSTRUCT LOCAL FILENAME

Problem
-------

mysqlbinlog client program is using default temporary directory for
constructing temporary local files.

Fix
---

Added a mysqlbinlog client program option into MTR configuration factory
to use the test case temporary directory for constructing temporary
local files.

* Bug 25364178 - XA PREPARE INCONSISTENT WITH XTRABACKUP

Issue: XA PREPARE, XA ROLLBACK and XA COMMIT for transaction
from a disconnected session does not take global commit lock
and modifies binlog and innodb redo log even when server is
in FTWRL. This could lead to inconsistent backups when
backup tools assume the server is in read-only state during
FTWRL mode.

Fix: Take global commit lock during these operations. Failure
to acquire lock during XA PREPARE will result in transaction
rollback. Failure to acquire lock during XA COMMIT and XA
ROLLBACK will return an error without changing transaction state.

* Bug#24748639:CAN'T SET SLAVE_SKIP_ERRORS > 3000

Problem:
MySQL 5.7 has introduced error numbers > 3000 but "slave_skip_errors"
doesn't allow error numbers which is greater than
3000(error numbers between 2000 and 2999 are for client side errors)

Fix:
Modify the MAX_SLAVE_ERROR variable, and add check for the client
side error while checking for errors to be skipped.

* Bug#24352163: CANNOT COMPILE WITH CLANG ON MAC OS X 10.10.5

Fix "cannot take the address of an rvalue of type" compilation
error when compiling with -std=c++11.

Partial backport from trunk of:
ea56f33fc55183b88a8f5f6fd9e1abbf1176c9f6

Also similar to contribution from Davi Arnaut.

* Bug#25811814	DEBIAN/UBUNTU PACKAGES SHOULD BUILD WITH --FAIL-MISSING

* Changed --list-missing to --fail-missing
* Added exclusions for files that should not be included
* Added missing manpages to packages

* BUG#25681977 GROUP REPLICATION NODE ON OPC FAILED TO REJOIN AFTER REBOOT

PROBLEM
-------
For many maintenance operations, VMs on OPC need to be restarted, but
after a VM restart, the node fails to rejoin the GR group. The problem
is that the private IP is changed after a VM restart and the remaining
nodes have a cached reference to the previous private IP.

When a node reconnects, a node's description called "server" is reused
if there is any. The server contains a reference to the addrinfo
structure and the current implementation does not check whether the
stored information is still valid or not.

FIX
---
To fix this issue, a new addrinfo structure is always fetched while
adding a new node or connecting to it.

* final

* Bug #25361251: INCORRECT BEHAVIOR WITH INSERT ON DUPLICATE
               KEY IN SP

Issue 1:
--------
The VALUES part of the ON DUPLICATE KEY UPDATE can refer to
the column list of an INSERT statement. When a BLOB value
is used in such a clause as part of an INSERT statement in
stored procedure, this can result in incorrect behavior.

BLOB values require some special handling because of
problems like the one mentioned in Bug#13901905. This is
done using the "insert_update_values_map" member of LEX.
Field_blob values are added to this map during
Item_insert_value::fix_fields. But the cleaning up takes
place at the end of the statement in lex_end.

This works fine for conventional execution statements. But
for statements in the body of an SP, lex_end is called only
once per statement. This is irrespective of whether the
statement is executed in a loop. This can result in
incorrect behavior because "insert_update_values_map" needs
to be cleaned up after statement execute.

Solution 1:
------------
The solution is to make sure that lex->clear_values_map is
called at the end of statement execution.

Issue 2 (from Bug #24538207):
-----------------------------
The problem here is similar to the one mentioned above. The
INSERT is a prepared statement. Each PREPARE also creates a
new "insert_update_values_map" for handling the BLOB values.
Only lex_end cleans up here too.

This results in the same problem as above.

Solution 2:
------------
The "EXECUTE ..." part will be fixed by Solution 1. But for
the "PREPARE ..." statements, lex->clear_values_map will be
added at the end of "Prepared_statement::prepare".

* BUG#25250768: WRITING ON A READ_ONLY=ON SERVER WITHOUT SUPER
              PRIVILEGE.

Backport from mysql-5.7 to mysql-5.5 and mysql-5.6.

BUG#13969578: TEMPORARY TABLE IN A DATABASE ON A READ-ONLY
                  INSTANCE CAN BE OVERWRITTEN

Analysis:
========

Creation or modification of a persistent table by a non-super user
is NOT ALLOWED in read_only mode. Only TEMPORARY tables are allowed
to be created or modified in read_only mode. But the creation of
a persistent table was being allowed when a temporary table of
the same name existed.

The routine which denies updating a non-temporary table in a
read_only mode does not handle the case of creation of a regular
table when a temporary table of the same exists.

Fix:
===
Handled the condition where an attempt is made to create a persistent
table having the same name as that of the temporary table. Hence
the creation of a persistent table by a non-super user when a
temporary table of the same exists is denied under read_only mode.

* WL 5847# Fixed NDB build issue.

* BUG#25503458 - SOME CLASS B PRIVATE ADDRESS IS NOT PERMITTED AUTOMATICALLY

Description:
---------------
According to
https://dev.mysql.com/doc/refman/5.7/en/group-replication-options.html#sysvar_
group_replication_ip_whitelist, when group_replication_ip_whitelist = AUTOMATIC,
private network is permitted automatically.

Below 172.21.134.27/23 is class C private IP address, but 172.21.134.0 is not
permitted. How To Repeat? Set class C private address other than 172.16.X.X

Analysis:
----------------
Looking at the code in get_ipv4_local_private_addresses in gcs_xcom_networking.cc
we notice that local class B addresses are not being tested in a range but only
for 172.16.X.X

Suggested Fix:
----------------
Refactored get_ipv4_local_private_addresses to make numeric and range comparisons.

* Bug #24323288: MAIN.MYSQLBINLOG_DEBUG FAILS WITH A LEAKSANITIZER ERROR

Issue:
------
On a system with AddressSanitizer support that includes Leaksanitizer,
the test mysqlbinlog_debug fails.test when compiled with -DWITH_ASAN and
-DCMAKE_BUILD_TYPE=Debug option.

Cause:
------
When the events are read from remote server using mysqlbinlog utility,
in two scenarios the memory clean up is not done. Those scenarios are.

1) When fake Rotate_event occur. Places where
   Rotate_event->when.tv_sec == 0
2) When we are using raw mode with mysqlbinlog, then for those cases
   where the log file could not be created or write had some issues.

Solution for 5.6 and 5.7:
-------------------------
5.6:
===
Scenario (1) will be fixed by backport of Bug#21697461 from mysql-5.7,
which is to create a method that will handle the memory clean up and
call this method from all the suspected places.

Scenario (2) will be an extension to Bug#21697461, for cases discussed
above in the *cause* section.
5.7:
===
As we already have Bug#21697461 in mysql-5.7, scenario (1) need not be
addressed separately.
For scenario (2) same fix as above, extension of Bug#21697461.

Solution for trunk:
-------------------
Implemented a different solution(removed Bug#21697461) to fix scenario
(1) and (2).

- A new solution is implemented to make the area around the
- Manage the clean up of temp_buf through m_free_temp_buf_in_destructor
  variable added in Log_event class.
- Created a new RAII class Destroy_log_event_guard to handle memory of
  Log_event object created in the dump_remote_log_entries method.
- Extended mysqlbinlog_debug.test to test the changes done.

A new solution is implemented thinking that any modification done to
the area around the code causing memory leaks at present have lesser number
of memory issues in future.

* Bug #25549637: QUERY WITH GROUPBY AND HAVING GETS ERROR;
Bug #24497007: GROUPBY() AND HAVING() CLAUSES GENERATES WRONG SQL QUERY

Description:
In some cases of usage a crud.find message together with grouping
criteria the x plugin returns unproper results.

Fix:
Generation of sql query has been improved by moving 'grouping_criteria'
expression to the derived query.

Reviewed-by: Lukasz Kotula <lukasz.kotula@oracle.com>
RB:15532

* Bug #24754897 RPL_INJECTOR API FUNCTION FOR UPDATE_ROW WITH BEFORE AND AFTER COLUMN BITMAPS

An UPDATE_ROW binlog event contains 2 versions of the updated
row: a Before Image which has the row before the update,
and an After Image of the row after the update. The Before
and After Image do not always include all the columns in the
row. The binlog format decides which columns will be included.

There is a requirement in ndbcluster to reduce the size of the
binlog event while retaining the UPDATE_ROW format of
BeforeImage/AfterImage. This can be done by writing only
primary key columns from the Before Image and only modified
columns from the After Image. However, the existing binlog
formats do not support different sets of columns in the Before
and After Images.

The --binlog-format=MINIMAL option is not usable for ndbcluster
because enabling the binlog format code breaks the
--ndb-log-updated-only option in ndb replication.

This patch implements 2 new ndb binlog formats for UPDATE_ROW
where the Before Image and After Image are minimised to remove
duplicate information.

UPDATED_ONLY_USE_UPDATE_MINIMAL: log update as UPDATE_ROW, log
only primary key columns in Before Image and only updated columns
in After Image

FULL_USE_UPDATE_MINIMAL: log update as UPDATE_ROW, log only
primary key columns in Before Image and all non primary key
columns in After Image

This means that each column is contained only once in each event,
bringing efficiency close to that achieved by the variants logging
with WRITE_ROW events.

The UPDATED_ONLY_USE_UPDATE_MINIMAL and FULL_USE_UPDATE_MINIMAL
formats can be used with the conflict resolution algorithms
NDB$EPOCH(), NDB$EPOCH2() and NDB$EPOCH_TRANS(), because these
do not require the 'Before' values of non-primary-key-columns.
Conflict resolution algorithms like NDB$MAX() and NDB$OLD()
require the old values of non-PK columns, and will not work
correctly with these new variants.

* Bug #24753281: REQUEST FOR SLAVE_SKIP_ERRORS = ER_INCONSISTENT_ERROR

Problem:
--------
Replication applier thread returns 3002, ER_INCONSISTENT_ERROR, when there's
difference between expected error number and actual error number on slave.

So when ER_INCONSISTENT_ERROR is passed as an input to slave-skip-errors,
the replication should continue despite having different expected error and
actual error on slave.

But in current code, the slave is not able to ignore ER_INCONSISTENT_ERROR.

Fix:
----
Check if ER_INCONSISTENT_ERROR is passed in the list of errors to be
skipped. If yes then do not throw error, just log this info with the mention
of actual and expected error  as a note in the slave's error log.

* Bug#25043674 - MYSQLACCESS SCRIPT LOADS AND EXECUTES CODE
               FROM THE CURRENT DIRECTORY

DESCRIPTION
===========
When 'mysqlaccess' tool is run, it reads (and executes) the
content of its configuration file 'mysqlaccess.conf' from
the current directory. This is not a recommended behaviour
as someone with ill intentions can insert malicious
instructions into this file which could be executed
whenever this tool is run.

ANALYSIS
========
The configuration file is presently looked for, in the
following folders (in given order):
1. Current directory
2. SYSCONFDIR       //This gets expanded
3. /etc/

Owing to the reasons mentioned above, we should not permit
the file to be in the current directory. Since the other
two folders are assumed to be accessible only to authorized
people, the config file is safe to be read from there.

FIX
===
Modified the script so that it looks for the config file
now in the following two folders (in the given order):
1. SYSCONFDIR
2. /etc/

If it's absent from above locations but present in current
directory, an error is thrown asking the user to move the
file to one of the above locations and retry.

NOTE
====
The location paths and their precedence are not documented
for this tool. It needs to be noted as part of the
associated documentation.

* Bug#25876807: FOR DIFFERENT VERSION (5.7 AND 8.0) GROUP, LEADER SHOULD BE
              ONLY FROM LOWER VERSION

PROBLEM STATEMENT
=================
When a group contain members of two different versions (5.7 & 8.0), higher
version members are put in read compatability mode and cannot write to the
group. In single primary-mode, currently the primary member can be elected
from the higher version members also thus making a member what should be in
read mode writable, what can lead to failures in the lower version members.

PROPOSED SOLUTIONS
==================
The primary member should be elected from lower version members i.e 5.7, in
single primary-mode. And so only the members with the lowest versions should
participate in the election.

* Bug #25469161: MYSQLDUMP SHOULD EXCLUDE REPLICATION
               REPOSITORY METADATA TABLES

Description : Replication meta data tables such as
mysql.slave_master_info and mysql.slave_relay_log_info
gets backed up when MYSQLDUMP is executed.

Analysis:- The above mentioned replication tables should be
excluded from backups because they are not a logical part of
data. These tables contain replication channel meta data
that will leave the replication meta data in an invalid
state after the restoration of the dump.

Fix: Both the tables are excluded from MYSQLDUMP.

* Bug#25469190: MYSQLPUMP SHOULD EXCLUDE REPLICATION
              REPOSITORY METADATA TABLES

Description : Replication meta data tables such as
mysql.slave_master_info and mysql.slave_relay_log_info
gets backed up when MYSQLPUMP is executed.

Analysis: The above mentioned replication tables should be
excluded from backups because they are not a logical part of
data. These tables contain replication channel meta data
that will leave the replication meta data in an invalid
state after the restoration of the dump.

Fix: Both the tables are excluded from MYSQLPUMP.

* Bug#25876807: FOR DIFFERENT VERSION (5.7 AND 5.8) GROUP, LEADER SHOULD BE ONLY FROM 5.7

Updating missing @param lowest_version_end definition for
sort_members_for_election() and removing @param lowest_version_end
for sort_and_get_lowest_version_member_position() in gcs_event_handlers.h
for Group Replication plugin.

* Rename ndbd_redo_log_reader man page to ndb_redo_log_reader

* Bug #25331425: DISTINCT CLAUSE DOES NOT WORK IN
               GROUP_CONCAT

Issue:
------
The problem occurs when:
1) GROUP_CONCAT (DISTINCT ....) is used in the query.
2) Data size greater than value of system variable:
tmp_table_size.

The result would contain values that are non-unique.

Root cause:
-----------
An in-memory structure is used to filter out non-unique
values. When the data size exceeds tmp_table_size, the
overflow is written to disk as a separate file. The
expectation here is that when all such files are merged,
the full set of unique values can be obtained.

But the Item_func_group_concat::add function is in a bit of
hurry. Even as it is adding values to the tree, it wants to
decide if a value is unique and write it to the result
buffer. This works fine if the configured maximum size is
greater than the size of the data. But since tmp_table_size
is set to a low value, the size of the tree is smaller and
hence requires the creation of multiple copies on disk.

Item_func_group_concat currently has no mechanism to merge
all the copies on disk and then generate the result. This
results in duplicate values.

Solution:
---------
In case of the DISTINCT clause, don't write to the result
buffer immediately. Do the merge and only then put the
unique values in the result buffer. This has be done in
Item_func_group_concat::val_str.

Note regarding result file changes:
-----------------------------------
Earlier when a unique value was seen in
Item_func_group_concat::add, it was dumped to the output.
So result is in the order stored in SE. But with this fix,
we wait until all the data is read and the final set of
unique values are written to output buffer. So the data
appears in the sorted order.

* Bug#23280117: 5.7 BUG XXXXX TEST REPLACE NUMBER ROUND+GIS PRECISION DIFFERENCES
              (CONTRIBUTION)

Issue:
------
MTR tests written to test GIS functions fail with a result content mismatch
on some platforms due to difference in precision. These have been corrected
using the mysqltest function --replace_regex till now.
This requires a complex regular expression to replace the result with a lower
precision.

Fix:
----
For ease of use, a new mysqltest command 'replace_numberic_round' has been
introduced to round off numeric values.
The usage is as follows:
--replace_numberic_round <precision>
where precision is the number of digits after the decimal that the result will
be rounded off to. The precision can be any number between 0 to 16.

Patch contributed by: Daniel Black
Reviewed-by: Pavan Naik <pavan.naik@oracle.com>
Reviewed-by: Srikanth BR <srikanth.b.r@oracle.com>
RB: 14633

* Bug#23280117: 5.7 BUG XXXXX TEST REPLACE NUMBER ROUND+GIS PRECISION DIFFERENCES
              (CONTRIBUTION)

Post push fix: Changing name of the mysqltest command from
'replace_numberic_round' to 'replace_numeric_round'.

Reviewed-by: Pavan Naik <pavan.naik@oracle.com>

* BUG #23107137: TYPOS IN MYSQLXTEST --HELP-COMMANDS

Description
===========
Using mysqlxtest with --command-help switch gives output containing some
minor typos. --help provides incomplete information for the --bindump help
entry.

RB: 16011
Reviewed by: Grzegorz Szwarc <grzegorz.szwarc@oracle.com>
Reviewed by: Lukasz Kotula <lukasz.kotula@oracle.com>

* ldap: Changed ldap sasl client library to dynamic library.

* Bug#25452277: OUT OF DISK SPACE CONDITION LEADS TO SIGNAL 11

Problem:
========
An out of space condition has been developed, after some
warnings mysqld exits abnormally with signal 11.

Analysis:
=========
User has configured binlog_error_action= ABORT_SERVER.
During master server execution disk becomes full and due to
this, update on binary log index file fails. On this critical
error server aborts. Now users try to restart the server but
still there is no free disk space. While the server is
trying to come up it finds the binary log being enabled it
tries to open the file and update the index file with this
information. Again it results in critical error and code
tries to access server thread specific diagnostic area which
is not yet initialized. This causes the server to exit
abnormally.

Fix:
===
In existing code when server encounters any critical error
while it is coming up it just logs appropriate error message
into the error log and aborts. Similarly, binlog error exit
handling code should just log message into error log when
the actual 'thd' object is not yet ready.

Write error message into error log and exit.

* Bug#25912557 ST_LATFROMGEOHASH/ST_LONGFROMGEOHASH ODD INPUT TYPE REQUIREMENT

The functions ST_LatFromGeohash, ST_LongFromGeohash and ST_PointFromGeohash
did not allow data that originated from a CHAR column.

Change-Id: I88f7fed5ad992394c18a945f00896c033b0688ba

* BUG#25800025: SQL THREAD ERROR 1778 WHEN GTID SKIPPING PARTIAL TRX

Problem
-------

SQL thread is unable to GTID skip a partial transaction.

Analysis
--------

The SQL thread has a mechanism to rollback the partial transactions left
on the relay log. The mechanism is activated every time the SQL thread
is going to apply a Gtid_log_event and the THD already owns a GTID.

The issue is happening when the transaction being applied is to be GTID
skipped. Because in this case the THD owns no GTID, the rollback
mechanism is not being used.

Fix
---

When applying a Gtid_log_event, the SQL thread will "rollback" the
partial transaction when a transaction that was being skipped was
incomplete in the relay log.

* Bug #25264253: QUERY REWRITER PLUGIN CRASHES WHEN READ_ONLY IS
               ENABLED

ISSUE: When server is started with read_only option enabled, query
       rewriter plugin fails to take appropriate lock resulting in
       assertion and server exiting abnormally.

FIX: Made change to ignore read_only option so that the query rewiter
     plugin takes appropriate lock.

* BUG#25476479:  MYSQLD FILLS UP ERROR LOGS WITH [ERROR]
               ERROR IN ACCEPT: BAD FILE DESCRIPTOR

DESCRIPTION:
============
When mysqld is secured with TCP wrappers and the files
hosts.allow and hosts.deny are configured to restrict access
from an IP address, very large error log files are generated
if we try to connect to the server via that IP address.
This runs indefinitely resulting in ever growing error log
files and puts load on the machine. The server continues to
serve authorized connections although slowly.

ANALYSIS:
=========
When the listening socket is accepting connections, if a
connection was refused by TCP wrappers, we were closing the
listening socket which led to connect socket being polled
continuously, leading to indefinite number of errors being
logged to the error log.

FIX:
====
We are now closing the connect socket only, thus avoiding
its polling and also allowing the listening socket to
accept other connections.

* Bug#25909965 REMOVE DEPENDENCIS ON LIBSTATOMIC.SO ON SOLARIS

man -M /opt/developerstudio12.5/man CC
       -xatomic=a
           Specify which atomics support runtime library is linked.
           a must be one of the following:
           studio    Link with the libstatomic  library  bundled  with  Oracle
                     Developer Studio.
           gcc       Link with the libatomic library in /usr/lib.
           none      Do not link with any atomics support library.
           The option -xatomic=studio is the default when neither -latomic nor
           -xatomic  is  specified,  and  the   compiler   is   compiling   in
           -std={c++03|c++11|c++14} mode.

So with -latomic we got dependency on
/opt/developerstudio12.5/lib/compilers/atomic/sparcv9/libstatomic.so

For InnoDB memory barriers: prefer HAVE_IB_MACHINE_BARRIER_SOLARIS
over HAVE_IB_GCC_ATOMIC_THREAD_FENCE on Solaris, since the latter is
really part of C11/C++11 and creates a dependency on libstatomic which
is not installed by default.

* Bug#25731075: INCOMPATIBLE POINTER TYPES WARNING WHEN BUILDING XCOM

Broken build for gcc/Solaris
Backport from trunk.

* Added plugins for commercial docker rpm builds

* Bug #25793677   INNODB: FAILING ASSERTION: CLUST_TEMPL_FOR_SEC || LEN

Problem:
========
During searching, length of the virtual column field from virtual index record is
less than the (expected) template column length and it leads to failure.

Fix:
===
Length of the virtual column field from virtual index record can be less than
template column length if record is fetched from prefixed virtual index.

Reviewed-by: Jimmy Yang jimmy.yang@oracle.com
RB: 16047

* WL10166 Result missmatch content fix

* BUG#25800025: SQL THREAD ERROR 1778 WHEN GTID SKIPPING PARTIAL TRX

Post-push fix for non-deterministic behavior of the MTR include file
'rpl_receive_event_count.inc' sourced by
'rpl_nogtid_rollback_on_anonymous_gtid' test case.

* Bug#25942414: SSL VARIABLES USAGE WHEN LIBMYSQL IS COMPILED WITH WITH_SSL=NO

Description: If libmysql is compiled with WITH_SSL=NO,
             --ssl-* are not useful.

Solution: 1. Restricted WITH_SSL to values : bundled | yes | system
          2. Made "bundled" as default value for WITH_SSL. Also,
             not specifying WITH_SSL or even specifying WITH_SSL=no
             will be treated as/converted to WITH_SSL=bundled.

Reviewed-By: Tor Didriksen <tor.didriksen@oracle.com>
Reviewed-By: Georgi Kodinov <georgi.kodinov@oracle.com>

* Raise version number after cloning 5.5.55

(cherry picked from commit 7df8dc750c26ead87c643f87dccba28a66cf3c9b)

* Bug#25942414: SSL VARIABLES USAGE WHEN LIBMYSQL IS COMPILED WITH WITH_SSL=NO

Description: If libmysql is compiled with WITH_SSL=NO,
             --ssl-* are not useful.

Solution: 1. Restricted WITH_SSL to values : bundled | yes | system
          2. Made "bundled" as default value for WITH_SSL. Also,
             not specifying WITH_SSL or even specifying WITH_SSL=no
             will be treated as/converted to WITH_SSL=bundled.

Reviewed-By: Tor Didriksen <tor.didriksen@oracle.com>
Reviewed-By: Georgi Kodinov <georgi.kodinov@oracle.com>
(cherry picked from commit 3eb2058be34d1a21771fe89ff1a0c08f156899bc)

* Raise version number after cloning 5.5.56

* Bug #23573575: GCOLS: INCORRECT BEHAVIOR WITH BLOB IN DML

Issue:
------
This problem occurs when:
1) A virtual generated column is of type BLOB and is part
   of a secondary index.
2) A REPLACE statement tries to update an existing row.

The row in the clustered index has the new value but the
value of the virtual generated column in the secondary
index has not been updated.

This can result in incorrect behavior with future DML
statements.

Root cause:
-----------
ha_index_read_map calls update_generated_read_fields.
Field_blob object is required here to keep track of the
actual value of column 'c'.

                       a   b            c
table->record[1] ==> --------------------------------
                     | 7 | 7  |   ptr_to_blob_value |
                     --------------------------------
                                        ||
                                        \/
                                      ......................
                                      | <to-be-calculated> |
                                      ......................

                       a   b            c
table->record[0] ==> --------------------------------
                     | 7 | 13 |   ptr_to_blob_value |
                     --------------------------------
                                        ||
                                        \/
                                      ........
                                      |  13  |
                                      ........

table->record[0] was calculated in update_generated_write_fields.

update_generated_read_fields tries to re-create the
existing row to see if the index's value needs
modification. Here it has to use the same blob object
referred to above.

At the end of update_generated_read_fields :
                       a   b            c
table->record[1] ==> --------------------------------
                     | 7 | 7  |   ptr_to_blob_value |
                     --------------------------------
                                        ||
                                        \/
                                      ........
                                      |  7   |
                                      ........
                                        /\
                                        ||
                       a   b            c
table->record[0] ==> --------------------------------
                     | 7 | 13 |   ptr_to_blob_value |
                     --------------------------------

update_generated_read_fields never calls
Field_blob::keep_old_value and hence the value is
over-written with the current row. This allows the server
to mistake that no change is required for the secondary
index. This results in a mis-match between the values in
the clustered index and the ones in the secondary index.

Solution:
---------
Effectively what we need is a producer-consumer type model:
Whichever function (among update_generated_write_fields and
update_generated_read_fields) is called first should set
the m_keep_old_value flag to true.

The succeeding function (again among the two) should consume
the flag. If the flag isn't consumed, at the end of the
statement, set it to false. This will avoid issues that
arise when, for example an INSERT preceeds an UPDATE.

There are other exceptions, like
update_generated_write_fields being called from
COPY_INFO::set_function_defaults. This was done as part of
Bug#21875520, to avoid a much larger re-factoring. The flag
is set to false in this case.

* Bug#25287633:GCOLS: INCORRECT BEHAVIOR WITH CHARSET CHANGES

Issue:
------
Virtual generated column functions that contain certain
string functions can cause problems when there is a change
in charset.

Steps:
1) Create table t1 with virtual generated column containing
string functions that generate strings in the connection
character set.

2) SELECT * FROM t1;
This step creates the item tree for the generated column.
While creating the objects, the table's memroot is used.
The flag "permanent_changes_completed" is set to true to
indicate that no further permanent changes will take place
in the item tree.

3) SET NAMES utf8;
Changes the connection's default charset to utf8.

4) SELECT * FROM t1;
While re-fixing the generated column items
(TABLE::refix_gc_items), it is noticed that the default
charset of the result of some string functions doesn't
match with other parameters. So a new Item object is
created with the correct charset. This new object now
replaces the one present in the item tree (created in
step 2).

But the new object is created on the statement's
mem_root and hence will be destroyed at the end of the
statement. Thus leaving the generated column's item tree
with a dangling pointer.

5) SELECT * FROM t1;
While going through the generated column's item tree,
dangling pointer strikes.

The root cause here is that a conversion due to charsets is
required and a new Item creation is required in step 4.
This should be avoided.

Solution:
--------
In the resolve_type phase, some string functions assign the
current charset to the Item object. This can return strings
in the connection character set.

Use the table's charset (mentioned in the DDL) while doing
fix_fields. Before fix_fields is called on the generated
column's expression, switch to the table's charset. This
will side-step any problem with aggregating charsets since
it is independent of any charset changes made by the user
after the DDL.

* BUG#25719975 SHEBANG HARD CODED AS /USR/BIN/PERL IN SCRIPTS, BREAKS ON FREEBSD

Post push fix for Solaris.

* Bug #25943038: QUERY WITH DERIVED TABLE IS CACHED IN MVCC
               SCENARIO

Issue:
------
Queries with derived tables which used materialization were
stored in Query Cache without fully taking into account storage
engines for some tables used by them (specifically information
for underlying tables of derived table was disregarded).
So, cached results for such queries were used without check
against SEs which led to stale results sometimes.

This problem occurs when:
1) Rows are being inserted into a table (t1 for example)
   but they haven't been committed yet. Storage engine
   used is INNODB
2) Query with a derived table contains t1 as a base table.
3) Optimizer has chosen to materialize the derived table.

In this case Query Cache returned stale results for the query
after transaction commit.

Solution:
---------
When a derived table is present, no check is performed to
make sure that all the base tables qualify to be cached.
This has been corrected. If the query qualifies to be
cached, then the storage engine will be consulted at cache
lookup time.

* Bug#25984429 MTR: BACKPORT OPTION --report-unstable-tests TO 5.7

Problem:
--------
Various Linux distributions are running the MySQL Test Suite during
build. As known, MySQL test suite has some flaky tests. If a build
is unlucky and gets a sporadic test failure, it is cancelled and
in the worst case, the distribution is blocked from upgrading MySQL
to a new one. Even if MTR is run with --retry > 1 and all failed
tests pass on retry, exit code from MTR is non zero and build stops.

Solution:
---------
An MTR option named --report-unstable-tests is added which does the
following:
- Reports any test which has passed on at least one retry attempt
  in a separate category called "Unstable tests" in the summary.
  They will be in the format :
  <test_name>(<number_of_failures>/<number_of_attempts>)
- In case all failures are due to unstable tests, it causes MTR
  to print a warning and terminate with a zero exit code.

Note:
This is a backport of the patch for Bug#24473420 from mysql-8.0

Reviewed-by: Pavan Naik <pavan.naik@oracle.com>

* Bug #24929748	REVERSE SCAN ON A PARTITIONED TABLE DOES ICP
              CHECK INCORRECTLY, CAUSING SLOWDOWN

PROBLEM

Partition helper functions assume that end range condition will
be checked by the server and reset the end range condition
causing innodb to do scans beyond the end range condition passed
by ICP from optimizer.

FIX

Remove the redundant resetting of end ranges in partition
helper functions

[ #rb 15977 approved by deb ]

* Bug#25340722 - PRINT BINARY DATA AS HEX IN THE MYSQL
               CLIENT (CONTRIBUTION)

DESCRIPTION:
============
Binary data should be printed as hex in the mysql client
when the option binary-as-hex is enabled.

ANALYSIS:
=========
The fix deals only with mysql command line client.
It does not change, at all, the data sent to the
applications. Printing binary data as hex also
allows to use the output in the where clause
of the query.

FIX:
====
A new option 'binary-as-hex' is introduced to print the
binary contents as hex in the mysql client. The option
is disabled by default. When the option is enabled, we
convert the binary data to hex before printing the
contents irrespective of whether it is in tabular,
xml or html format.

* Bug#25998285 - ADD MYSQLADMIN DOCKER/MINIMAL SERVER BUILDS

- mysqladmin is needed by InnoDB Cluster, add tool to docker/minimal package.

* BUG#25385334: MYSQL 5.7 ERROR WITH ALTER TABLE MODIFY SYNTAX
              AND DATETIME TYPE.

Analysis:
========
The problem reported in the bug is:
ALTER TABLE MODIFY operation with FIRST/AFTER clause for
DATE/DATETIME and GEOMETRIC types to modify the position
of columns failed with an error.

ALTER TABLE operations for DATE/DATETIME and GEOMETRIC
types require additional handling. The cases where
an error should be reported is discussed below:

a) When a new NOT NULL column of type DATE/DATETIME/GEOMETRIC
   without a default value is added to a table which is not
   empty, a zero date and empty string is used. Under
   NO_ZERO_DATE and STRICT SQL mode, zero date is not allowed.
   An empty string is an invalid value for GEOMETRIC type.
b) When a column of different data type which can hold NULL
   value is changed/modified to GEOMETRIC NOT NULL type,
   the NULL value is converted to a empty string which is
   invalid value for GEOMETRIC types. Note that data type
   conversion of existing columns is possible only using
   COPY algorithm. Attempt to change/modify datatypes of
   existing columns using INPLACE algorithm is not supported
   and will fail with an error.
c) GEOMETRIC/DATE/DATETIME can be modified from NULLABLE
   columns to NOT NULL using COPY and INPLACE algorithm.
   When the table holds NULL value, an error is reported
   at SQL layer for COPY algorithm, while the SE enforces
   the restriction for INPLACE algorithm.

Patch for bug19880316 utilized the flag 'error_if_not_empty'
to enforce the restrictions for GEOMETRIC types listed above
for COPY algorithm and also handled the case of ALTER TABLE,
ADD column for INPLACE algorithm.

The patch for bug19880316 introduced a bug-ALTER TABLE MODIFY/
CHANGE operation with FIRST/AFTER clause for GEOMETRIC types
using COPY/INPLACE algorithm to modify the position of columns
failed with an error. The scope of the bug was extended to
DATE/DATETIME types with the bug fix for bug16888677 which is
now raised in the current bug report.

The flag 'error_if_not_empty' is set in 'mysql_prepare_alter_table()'
when:
* Field is DATE/DATETIME or GEOMETRIC.
* Field is NOT NULL and DEFAULT value is not supplied.

The flag is later utilized to report error "Invalid use of NULL
value" in mysql_inplace_alter_table() if:
* The MDL lock is not upgraded from SU to block writes.
* Table is not empty.

The ALTER operation reported in the bug report can be performed
using INPLACE algorithm which does not require the SE to block writes.
Hence the MDL lock is not upgraded. The flag was incorrectly set for
the above ALTER operation which triggered the error.

Fix:
====
a) The flag 'error_if_not_empty' is set when new NOT NULL columns
   of DATE/DATETIME/GEOMETRIC types are added.(Note that the other
   case where the flag is set when datatype is modified/changed to
   GEOMETRIC NOT NULL from NON-GEOMETRIC types using COPY algorithm
   introduced by patch for bug 19880316).

b) Also the check to ensure that MDL lock is upgraded
   from SU is removed from 'mysql_inplace_alter_table()'.
   ALTER TABLE, INPLACE algorithm does not support
   changing/modifying data type. So, the flag
   'error_if_not_empty' is set only for ALTER TABLE,
   ADD which takes an exclusive lock.

c) ALTER TABLE, MODIFY/CHANGE using INPLACE algorithm to change
   columns from NULL to NOT NULL, the validation is pushed down
   to SE.

d) Change in error message is recorded for the test
   'alter_spatial_index.test' in innodb_gis since the fix
   pushes the check during the data validation.

* Bug #24960450   CONCURRENT DELETE DOESN'T APPLY DURING TABLE REBUILD

Analysis:
========
During alter table rebuild, InnoDB fails to apply concurrent delete log.
Parsing and validation of merge record happens while applying the
log operation on a table. Validation goes wrong for the virtual column.
Validation assumes that virtual column information can't be the end
of the merge record end.

Fix:
====
Virtual column information in the merge record can be end of the merge
record. Virtual column information is written at the end for
row_log_table_delete().

Reviewed-by: Satya Bodapati<satya.bodapati@oracle.com>
RB: 16155

* Bug #25814143: ADD SUPPORT FOR RUNNING UNITTESTS FROM SPEC FILES

Add separate setting to run unittest while building RPM.

* Bug #25814143: ADD SUPPORT FOR RUNNING UNITTESTS FROM SPEC FILES

Don't fail build when unittests fail.

* Bug#24559588	CANNOT RUN MULTIPLE MYSQL INSTANCES UNDER SYSTEMD

Added new service file for mysql@.. services. This service is disabled by default.
Updated systemd scripts to take instance info when checking config, etc.

Known issues:
* User needs to manually update/disable apparmor profile for new data locations
* mysql@-instances must be manually stopped before uninstallation or upgrade

* WL#10166 test fix

* WL #5847: LDAP, fixed sasl test cases and some cleanup.

* WL#10166 Require WKB input in WKB parsing functions

Post push fix for failing NDB testcase.

* BUG#25793366: GR NODE IS IN RECOVERING STATE IF BINLOG_CHECKSUM CONFIGURED ON RUNNING SERVER

When node configured in GR setup with binlog_checksum='NONE' on a running
server, then start group replication return success but node is in RECOVERING
mode forever and server shutdown hangs at this moment. Only option is kill the
server.

The Continuation object used on the plugin applier pipeline to wait for the
applications of events was the source of the issue.  Multiple invocations of
the waiting for execution method when there was a previous error lead to a
stuck pipeline.

As a solution the waiting method no longer waits when there is a known previous
error.

* BUG#25451091:CREATE TABLE DATA DIRECTORY / INDEX DIRECTORY
             SYMLINK CHECK RACE CONDITIONS

ANALYSIS:
=========
A potential defect exists in the handling of CREATE
TABLE .. DATA DIRECTORY/ INDEX DIRECTORY which gives way to
the user to gain access to another user table or a system
table.

FIX:
====
The lstat and fstat output of the target files are now
stored which help in determining the identity of the target
files thus preventing the unauthorized access to other
files.

* BUG#25793366: GR NODE IS IN RECOVERING STATE IF BINLOG_CHECKSUM CONFIGURED ON RUNNING SERVER [Post-Fix]

Missing two mtr suppression on gr_binlog_checksum_failure test

* BUG#25040331: INTERLEAVED XA TRANSACTIONS MAY DEADLOCK SLAVE APPLIER
              WITH REPEATABLE READ

Problem and Analysis
--------------------

    Because of the use of MTS and/or XA transactions, the order statements
    from distinct transactions were originally applied on the master may not
    be the same order they are applied on the slave.

    On those cases, using RBR with a transaction isolation level other than
    READ COMMITTED might might lead to blocking conditions that didn't
    happened on the original workload (see BUG#25082593).

Fix
---

    We are now using a flag on the [Anonymous_gtid|Gtid]_log_event stating
    if the transaction might have changes logged with statements.

    When this flag is set (default for non-fixed servers) the slave will not
    change the applier thread transaction isolation level.

    When the flag is not set, the applier will set the transaction isolation
    level of the current transaction to READ COMMITTED if necessary (it will
    not change it when the applier session isolation level be equal or less
    restrictive than READ COMMITTED).

    In this way, we are forcing that all transactions replicated using row
    based replication will always use the READ COMMITTED transaction
    isolation level on the slave side, regardless of any server transaction
    isolation level configuration.

    The fix also restored the applier session transaction isolation level
    after applying a Xid_log_event (after committing a replicated
    transaction), a XA_prepare_log_event (after preparing a replicated
    transaction) and in the cases of aborted/failed transactions.

    The mysqlbinlog dump of a transaction flagged as "rbr_only" will also
    include a "SET TRANSACTION ISOLATION LEVEL" statement when dumping the
    GTID event.

Changed files:

    @ libbinlogevents/include/control_events.h

      We gave a meaning to the already logged flag information.

    @ libbinlogevents/src/control_events.cpp

      Processed the GTID flags, collecting the "might_have_sbr_stmts" info.

    @ plugin/replication_observers_example/replication_observers_example.cc
    @ rapid/plugin/group_replication/src/observer_trans.cc
    @ rapid/plugin/group_replication/src/handlers/certification_handler.cc

      Fixed calls to GTID constructor.

    @ sql/binlog.cc

      Created a new flag at "binlog_cache_data" named "with_sbr" that is
      set true when the cache is fed with non-terminal query event.

      Created a new flag at "binlog_cache_data" named "with_rbr" that is
      set true when the cache is fed with Table_map or Rows event.

      Created also the "might_have_sbr_stmts" function, that returns false
      for "pure RBR" transactions. This new function was implemented for
      the "binlog_cache_data".

      Added the logic to reset and set the new flags at
      "binlog_cache_data::reset" and "binlog_cache_data::write_event", and
      also to preserve the flags upon "SAVEPOINT/ROLLBACK TO SAVEPOINT".

      Made "MYSQL_BIN_LOG::write_gtid" to consider the cache information
      about the binary log content when creating the Gtid_log_event.

    @ sql/log_event.h

      Updated "Query_log_event::is_trans_keyword()" to recognize XA
      transactions terminal keywords that should not be considered as
      statements changing content.

      Created a new function "Log_event::is_sbr_logging_format()" that
      will return true if the event was/has to be logged using SBR and
      does not contains a transaction control statement.

      Created a new function "Log_event::is_rbr_logging_format()" that
      will return true if the event was/has to be logged using RBR.

    @ sql/log_event.cc

      We will stop the applier with an error when we assumed the
      transaction as "rbr_only" and some statement changing content be
      applied.

      Restored the applier session transaction isolation level
      after applying a Xid_log_event (after committing a replicated
      transaction), a XA_prepare_log_event (after preparing a replicated
      transaction) and in the cases of aborted/failed transactions.

      "Gtid_log_event::print" will now also print "rbr_only=[yes|no]" when
      dumping a Gtid_log_event with mysqlbinlog. It will also include a
      "SET TRANSACTION ISOLATION LEVEL READ COMMITTED" statement when
      dumping the GTID event with "rbr_only=yes".

      "Gtid_log_event::write_post_header_to_memory" now computes the GTID
      flags field instead of just writing "1" to it.

      "Gtid_log_event::do_apply_event" now set the transaction isolation
      level to READ COMMITTED when the "might_have_sbr_stmts" flag be false.
      Note that we don't change the isolation level for skipped GTID and
      when the isolation level be less restrictive than READ COMMITTED.

* Bug#21280753    MTS MUST BE ABLE HANDLE LARGE PACKETS WITH
SMALL SLAVE_PENDING_JOBS_SIZE_MAX

Problem: MTS is unable to process an event which is smaller than
slave_max_allowed_packet but bigger than slave_pending_jobs_size_max.

Analysis and Fix:
In the current code , slave_pending_jobs_size_max is hard limit i.e., any
packet greater than slave_pending_jobs_size_max is rejected by Coordinator.
Slave SQL thread will be stopped with error
ER_MTS_EVENT_BIGGER_PENDING_JOBS_SIZE_MAX. Now in this patch,
this limitation is changed to soft limit.

If there is an event whose size falls in between soft limit
(slave_pending_jobs_size_max) and the hard limit (slave_max_allowed_packet),
then we consider it as big event and MTS coordinator will wait till all
workers queue becomes empty. After it is empty, then this big event will be processed.
i.e., only one event of this size will be executed at any given time.
All the next events ( big or small) will wait in the coordinator for this big event to finish
it's execution.

IO thread has already having logic to protect event size with
slave_max_allowed_packet size (as the hard limit).  When IO thread is
downloading the event from the master, if the event size is greater than
slave_max_allowed_packet size, IO thread stop by throwing error
ER_NET_PACKET_TOO_LARGE. Hence there are no changes required
there.

* Bug#23289541: THE PROXIES_PRIV TABLE IS NOT REPLICATED

Description: PROXY grants are not properly replicated.
             This fix makes sure that query rewrite is
             done properly.

Reviewed-By: Georgi Kodinov <georgi.kodinov@oracle.com>
Reviewed-By: Luis Soares <luis.soares@oracle.com>

* Bug#24559588	CANNOT RUN MULTIPLE MYSQL INSTANCES UNDER SYSTEMD

Post-push fix: One of the help script functions has a syntax error

* Bug#26077226 DATABASE INITIALIZE FAILING IF SELINUX IS ENFORCING MODE ON EL6

Under SELinux mysqld can't read files under /tmp, switch directory to
/var/lib/mysql-files which have correct SELinux context.

* Bug#26036570: NEW PRIMARY IS NOT ELECTED BASED ON LOWEST ID IN
LEXICOGRAPHICAL ORDER

Issue:
The member having lowest server uuid is selected as primary leader in single primary mode election.
But that got reversed in Bug#25876807 push, so now its electing primary as member having highest server uuid.

Solution:
The sorting method comparator_group_member_uuid should sort in ascending
lexicographical order instead of descending lexicographical order.

* BUG#25209512: CURRENT_TIMESTAMP PRODUCES ZEROS IN TRIGGER

ANALYSIS:
=========
Zero date(0000-00-00 00:00:00) is generated by the
'BEFORE INSERT' trigger defined on a table with TIMESTAMP
or DATETIME field having the default as CURRENT_TIMESTAMP
or NOW().

Similar behavior is observed with LOAD INFILE queries and
also with BEFORE UPDATE triggers.

Whenever a 'BEFORE INSERT' trigger is defined on a table
and an insert happens on it, the trigger body is executed
first and then the new record is written into the table.
If the trigger body tries to utilize the default values
of the fields in the table, the values used will be
incorrect.

Brief working of 'AFTER INSERT' triggers:
1. Set function defaults for the table.
2. Write record into the table.
3. Process the trigger defined on it.
  3.1 Execute the trigger body.

Brief working of 'BEFORE INSERT' triggers:
1. Process triggers defined on the table.
  1.1 Execute the trigger body.
2. Set function defaults for the table.
3. Write record into the table.

Hence when a 'BEFORE INSERT' trigger is defined for a
table, the evaluation of default functions is not done for
the table before processing the trigger. Thus the default
values of the table are not updated and if the trigger body
tries to utilize these values, incorrect result will be
obtained.

FIX:
====
We are now evaluating the function defaults for a table
before processing the 'BEFORE INSERT' or 'BEFORE UPDATE'
triggers defined on it.

* Bug#21280753    MTS MUST BE ABLE HANDLE LARGE PACKETS WITH
SMALL SLAVE_PENDING_JOBS_SIZE_MAX

Post push fix (Test designed in mysql-5.6 where we have only
slave_parallel_type=database type). Including
that restriction in the test.

* Bug #25332330	COUNT(*) ON INNODB SOMETIMES RETURNS 0

PROBLEM

When using Index merge optimization current partitioning code in 5.6
does a memcpy instead of column of copy of the columns on which index
read is done . Because of which wrong record gets copied and optimizer
thinks there is no matching record as per the search criteria returning
0 for the count(*) value . The problem is sporiadic as the plan is not
always using intersection .

SOLUTION

Added a similar column copy function that would do the copy of required
column and not overwrite the other as in 5.7.

Reviewed-by: Debarun Banerjee<debarun.banerjee@oracle.com>
RB: 16157

* BUG#25209512: CURRENT_TIMESTAMP PRODUCES ZEROS IN TRIGGER

Post push fix for test failure.

* Bug #25436469: BUILDS ARE NOT REPRODUCIBLE

Backport to 5.5

Current MySQL builds, even on Pushbuild, are not reproducible; they return
different results depending on which directory they are built from (and
Pushbuild uses several different directories). This is because absolute paths
leak into debug information, and even worse, __FILE__. The latter moves code
around enough that we've actually seen sysbench changes on the order of 4% in
some tests.

CMake seemingly insists on using absolute paths, but we can insert our own
layer between CMake and GCC to relativize all paths. Also give the right flags
to get debug information reproducible and turn off build stamping. This makes
the mysqld build 100% bit-for-bit reproducible between runs on my machine,
even when run from different directories.

* Bug#16212207 - LOAD XML INFILE PERFORMANCE WITH INDENTED
               XML

DESCRIPTION
===========
LOAD XML INFILE performance becomes painfully slow if the
tags' value has any space(s) in between them. They're
usually kept intentionally for indentation purpose.

ANALYSIS
========
The extra spaces are calling clear_level() many a times
which is having overhead of clearing taglist etc. This can
be avoided altogether by skipping all such spaces.

FIX
===
Trim all the starting whitespaces from the value before
passing it to read_value()

* Bug #22834591: unused-local-typedef warnings from boost 1.59

Description:
Boost version 1.59 produces lots of unused-local-typedef warnings.
Boost version 1.64 (used in version 8) does not.

Copying from client/dump/CMakeList.txt I add the compiler flag to ignore
these warnings.

Testing Done:
Compilation on linux:
Scanning dependencies of target keyring_file
[ 51%] Building CXX object plugin/keyring/CMakeFiles/keyring_file.dir/common/keyring_key.cc.o
[ 51%] Building CXX object plugin/keyring/CMakeFiles/keyring_file.dir/common/keys_container.cc.o
[ 51%] Building CXX object plugin/keyring/CMakeFiles/keyring_file.dir/common/keyring_impl.cc.o
[ 51%] Building CXX object plugin/keyring/CMakeFiles/keyring_file.dir/keyring.cc.o
[ 51%] Building CXX object plugin/keyring/CMakeFiles/keyring_file.dir/hash_to_buffer_serializer.cc.o
[ 51%] Building CXX object plugin/keyring/CMakeFiles/keyring_file.dir/buffered_file_io.cc.o
[ 51%] Building CXX object plugin/keyring/CMakeFiles/keyring_file.dir/file_io.cc.o
[ 51%] Building CXX object plugin/keyring/CMakeFiles/keyring_file.dir/buffer.cc.o
[ 51%] Linking CXX shared module keyring_file.so
[ 51%] Built target keyring_file
[ 51%] Built target validate_password

Compilation on Windows succeeds too.
mtr on Linux and Windows

RB:16241

* BUG#25420001 KEYRING_OKV PLUGIN FAILED TO WORK IF SELINUX IN ENFORCING MODE ON EL PLATFORMS

Correct SELinux context must be set on all files in keyring directory,
add recursive option to enable this.

* Bug#24829050 INDEX_MERGE_INTERSECTION OPTIMIZATION CAUSES
             WRONG QUERY RESULTS

Analysis
========
When using index_merge for InnoDB tables with a primary key,
the server does not set read_set correctly for index scan after
optimization. This does not cause a problem in other cases
because InnoDB does create templates during first call to
index_read(). Innodb uses sql_stat_start to decide initializing
templates in index_read(). sql_stat_start is set to false at
handler creation and set to true during first call to index_read().
In case of sub-query, during optimization we scan the index
and reset it for index read for execution. Further call to
index_read() does uses the templates created at reset as
inited is set to true during optimize phase.

This leads to InnoDB not reading primary key of the rows and
furthur leads to mismatch during index merge.

Fix
====
Provide correct read_set for each index during index_reset
to innodb for creating the right column template for reading.

* BUG#25209512: CURRENT_TIMESTAMP PRODUCES ZEROS IN TRIGGER

Post push fix:
Valgrind reported errors because INSERT and LOAD operations
compared the old data record (TABLE::record[1]) and the
new data record (TABLE::record[0]). New data record is
initialized only for UPDATE operations.

A check is now introduced to ensure only UPDATE operations
compare records.

* WL#9053: Group Replication: Push Group Replication Plugin to mysql-trunk

Build with gcc (GCC) 5.4.0
error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
 #define number_is_valid_port(n) ((n) > 0 && (n) <= UINT16_MAX)

Fix: cast UINT16_MAX to int
(cherry picked from commit 940e828bb8213624aea163d45ada65871c15ea12)

* Bug #24961167	CONCURRENT INSERT FAILS IF TABLE DOES REBUILD

Analysis:
=========
   During alter table rebuild, InnoDB fails to apply concurrent insert log.
If the insert log record is present across the blocks then apply phase
trying to access the next block without fetching it.

Fix:
====
During virtual column parsing, check whether the record is present
across the blocks before accessing the virtual column information.

Reviewed-by: Jimmy Yang <jimmy.yang@oracle.com>
RB: 16243

* Bug #25357789   INNODB: LATCH ORDER VIOLATION DURING TRUNCATE TABLE
                        IF INNODB_SYNC_DEBUG ENABLED

Analysis:
========

(1) During TRUNCATE of file_per_table tablespace, dict_operation_lock is
released before eviction of dirty pages of a tablespace from the buffer
pool. After evic…


grooverdan:
Its really not possible to work out what the flattening of commits is actually trying to do.

PR #162, opened 2017-07-27
gis tests: minimise rounding

grooverdan:
With the advent of https://github.com/mysql/mysql-server/blob/5.7/cmake/floating_point.cmake, most architecture/compiler differences within the GIS test suites have been eliminated. As such the proliferation of replace_numeric_result/regex/result isn't as needed as it once was. This PR removes as much rounding as possible.

There are 8 remaining replace_numeric_result calls, each containing the maximum possible precision needed to pass test suites on x86_64, ppc64le and ARM64. Interesting it is x86_64 that produces a result that is different to the identical results in ppc64le/ARM64 however tracing the exact cause it beyond current bandwidth.

closes #158 where most of the direction/discussion was. Thanks @akopytov


grooverdan:

PR #165, opened 2017-08-06
fix cmd-line-client Segmentation fault

ghost:


grooverdan:
Given this transfers allocation from the stack to the heap, and TC_BUFSIZE is a #define constant, how is this correcting the problem?

What is an example of generating a segfault? What is backtrace?

Is running memset on the original buf sufficient to prevent the segfault? "// add to fix" is a useless comment and I recommend removing it.

ghost:
**It seems OK now.** ,I try to repeat it today, and it seems OK.
**Last year,** I built mysql 5.7 from source in CentOS7.
 (http://cdn.mysql.com//Downloads/MySQL-5.7/mysql-5.7.12.tar.gz). 
Some dependence(ncurses, boost, cmake, bison, python,etc) are also built from source code.
List of dependence:
cmake : https://cmake.org/files/v3.5/cmake-3.5.2.tar.gz
ncurses : http://ftp.gnu.org/gnu/ncurses/ncurses-6.0.tar.gz
bzip2 : http://www.bzip.org/1.0.6/bzip2-1.0.6.tar.gz
boost : https://sourceforge.net/projects/boost/files/boost/1.59.0/boost_1_59_0.tar.gz
m4 : http://ftp.gnu.org/gnu/m4/m4-latest.tar.gz
bison : http://ftp.gnu.org/gnu/bison/bison-3.0.tar.gz
python : https://www.python.org/ftp/python/2.7.5/Python-2.7.5.tar.xz

build cmd：
 cmake \
 -DCMAKE_INSTALL_PREFIX=$HOME/mysql \
 -DMYSQL-DATADIR=$HOME/mysql/data \
 -DSYSCONFDIR=$HOME/mysql/conf \
 -DMYSQL_UNIX_ADDR=$HOME/mysql/log/mysql.sock \
 -DMYSQL_TCP_PORT=3306 \
 -DEXTRA_CHARSETS=all \
 -DDEFAULT_CHARSET=utf8 \
 -DWITH_SSL:STRING=bundled \
 -DWITH_ZLIB:STRING=bundled \
 -DWITH_BOOST=$HOME/.local/include \
 -DCURSES_INCLUDE_PATH=$HOME/.local/include

Start the server (without any problem), then try to access server with the client:
mysql -uroot -p
after input root's passwd
leads to : 
        Segmentation fault


I would reopen the pull request until i could repeat it.

PR #168, opened 2017-08-22
Bug #87483 Accelerated AES with ARMv8 Crypto Extensions in yaSSL

guyuqi:
Support for 64-bit ARMv8 AES cryptography extensions in yaSSL taoCrypt.
The AES with ARMv8 NEON intrinsics will optimized the performance rather than uses table-based lookup.

Environment: Hislicon Taishan D03/D05 , Cavium ThunderX , AMD seattle (Softiron)	
OS: ubuntu16.04/Debian 8 ,  Centos

use yassl taocrypt benchmark(SOURCE_HOME/extra/yassl/taocrypt/benchmark/benchmark.cpp):


guyuqi:

PR #169, opened 2017-08-25
Bug #87543: dd_sdi-t.cc assumes 'char' to be signed

akopytov:
Declare 'x' as a signed char explicitly in the CharPromotion test.


akopytov:

PR #170, opened 2017-08-25
Bug #79144: No hardware CRC32 implementation for AArch64

akopytov:
This is a 8.0 version of the previously submitted patch, adding support
for hardware CRC32 calculation on AArch64 CPUs.


akopytov:

PR #174, opened 2017-10-24
Bug #88197 Accelerated SHA1/SHA256 with ARMv8 Crypto Extensions

guyuqi:
Support for 64-bit ARMv8 SHA1/SHA256 cryptography extensions in yaSSL taoCrypt.
The SHA1/SHA256 with ARMv8 NEON intrinsics will optimized the performance rather than uses table-based lookup.

Change-Id: I66378124fe09fcf98f40f83ffe6b662e231f9741
Signed-off-by: Yuqi Gu <yuqi.gu@arm.com>

### DIFF
--- a/extra/yassl/taocrypt/include/sha.hpp
+++ b/extra/yassl/taocrypt/include/sha.hpp
@@ -24,6 +24,18 @@
 
 #include "hash.hpp"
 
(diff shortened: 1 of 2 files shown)


guyuqi:

PR #177, opened 2017-11-18
Cosmetic: rename function set_unknow_error to correct set_unknown_error

sjmudd:
Fix a typo which leaves function names looking rather ugly.


sjmudd:

PR #178, opened 2017-11-20
Ashe dev mysql

sunashe:
when I want to add a new channel with GTID replication protocol  on an exists slave instance, I have to reset master to clear all GTID sets  in this slave instance to add this new  gtid sets(old()+new()uuid:sidn-sidm....). So,it is not very friendly. I want to add a channel  without truncate
 my binary log.


sunashe:
I have signed  the Oracle Contribution Agreement (OCA)

PR #181, opened 2017-11-28
Define constant following its comment.

bgrainger:
Moves the definition of `CLIENT_SSL_VERIFY_SERVER_CERT ` to immediately follow the comment describing it. This fixes the documentation at https://dev.mysql.com/doc/dev/mysql-server/latest/group__group__cs__capabilities__flags.html#ga4e7c0220984fdf1e231344f450cfd51e and addresses [bug 88208](https://bugs.mysql.com/bug.php?id=88208).

(Reopening #175 because it didn't pick up my OCA status.)


bgrainger:

PR #183, opened 2017-11-30
OpenBSD build fix

devnexen:


devnexen:

PR #184, opened 2017-12-05
Export zlib version in a variable

dveeden:
If [innodb_log_compressed_pages](https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_log_compressed_pages) is used one needs to be careful which zlib version in used when using InnoDB recovery. Among others this could impact backup/restore with a hot backup tool or a storage snapshot.

This patch exposes the zlib version.
Currently this is only exposed via:
```
storage/innobase/srv/srv0start.cc:      ib::info() << "Compressed tables use zlib " ZLIB_VERSION
```

Related: https://bugs.launchpad.net/percona-xtrabackup/+bug/1736139

### DIFF
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -123,6 +123,7 @@
 #include "sql/xa.h"
 #include "template_utils.h"              // pointer_cast
 #include "thr_lock.h"
+#include "zlib.h"
 
 #ifdef WITH_PERFSCHEMA_STORAGE_ENGINE
 #include "../storage/perfschema/pfs_server.h"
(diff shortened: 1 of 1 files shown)


dveeden:
This is what it does:
```
mysql> show global variables like 'version_compile_zlib';
+----------------------+--------+
| Variable_name        | Value  |
+----------------------+--------+
| version_compile_zlib | 1.2.11 |
+----------------------+--------+
1 row in set (0.01 sec)
```

dveeden:

PR #186, opened 2017-12-21
mysys: eval_num_suffix(,_ull} - support T,P,E extensions

grooverdan:
Because its so sad when innodb_buffer_pool_size=8T doesn't work.


grooverdan:

PR #191, opened 2018-01-12
Fix for Bug #89139

dveeden:
https://bugs.mysql.com/bug.php?id=89139


dveeden:

PR #192, opened 2018-01-14
Add test case for Bug #89224

dveeden:
https://bugs.mysql.com/bug.php?id=89224


dveeden:

PR #194, opened 2018-01-26
modify ut_print_buf_hex()

happieme:
The result is wrong. For string "123", the result should be "0x313233", but it's not.


happieme:
The 16 should be 4, there's no error with the 15,but the "0x0F" is more clear.

PR #196, opened 2018-02-04
Use native host check from OpenSSL

dveeden:
This results in support for Subject Alternative Name
[Bug #68052 | SSL Certificate Subject ALT Names with IPs not respected with --ssl-verify-serve](https://bugs.mysql.com/bug.php?id=68052)

To completely fix the above but a call to `X509_check_ip` would also be needed.

Note that [X509_check_host(3)](https://www.openssl.org/docs/manmaster/man3/X509_check_host.html) says:
"Applications are encouraged to use `X509_VERIFY_PARAM_set1_host()` rather than explicitly calling `X509_check_host(3)`."
Which is described on https://wiki.openssl.org/index.php/Hostname_validation

However to use `X509_VERIFY_PARAM_set1_host()` we need to do that just before
creating the connection. That should be done in `ssl_do()` which is called from
`sslconnect()`. But then `ssl_do()` needs to know the `ssl_mode` and `hostname`, which
it currently doesn't. Note that `ssl_verify_server_cert()` is called when the
connection is already created.

### DIFF
--- a/sql-common/client.cc
+++ b/sql-common/client.cc
@@ -108,6 +108,10 @@
 #define SOCKET_ERROR -1
 #endif
 
+#ifdef HAVE_OPENSSL
+#include <openssl/x509v3.h>
+#endif
+
 #include <mysql/client_plugin.h>
 #include <new>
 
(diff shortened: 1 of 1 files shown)


dveeden:

PR #198, opened 2018-02-07
Add optional release_extra rpm tag to distinguish rebuilt rpms

sjmudd:
If I download a src rpm built by Oracle and want to rebuild it with different options then I can do something like:

`$ rpmbuild --rebuild --define 'with_ssl system' mysql-community-5.7.21-1.el7.src.rpm`

However, doing this means the rebuilt rpms will have exactly the same name as the original one. It's not clear from the rpm name alone[1] that I have built this new rpm so there's a clear chance of confusion.

This patch provides for an optional `--define` argument, `release_extra`, that can be used when rebuilding rpms from the src rpm so that the new rpm will be named slightly differently.

e.g.

`$ rpmbuild --rebuild --define 'release_extra .sjm' --define 'with_ssl system' mysql-community-5.7.21-1.el7.src.rpm`

would create a binary filename of the form `mysql-community-server-5.7.20-1.sjm.el7.x86_64.rpm` which is easily distinguishable from the original rpms built by Oracle. The release version change is minimal so any existing software that sees this will only need minimal adjustments should it see these newly named rpms.

This argument would NOT be used by Oracle but only by any users who want to simplify the way rebuilt packages can be distinguished from the upstream originals.

[1] `rpm -qi mysql-community-server` would show you the build host etc.

### DIFF
--- a/packaging/rpm-docker/mysql.spec.in
+++ b/packaging/rpm-docker/mysql.spec.in
@@ -52,7 +52,7 @@ Name:           mysql%{?cluster:-cluster}-%{product_suffix}-minimal
 Summary:        A very fast and reliable SQL database server
 Group:          Applications/Databases
 Version:        @MYSQL_NO_DASH_VERSION@
-Release:        0.1%{?milestone:.%{milestone}}%{?commercial:.1}%{?dist}
+Release:        0.1%{?milestone:.%{milestone}}%{?commercial:.1}%{?release_extra}%{?dist}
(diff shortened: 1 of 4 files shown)


sjmudd:
Note: this PR is for 8.0. I'd like to do an equivalent one for the current GA version 5.7 too.

sjmudd:

sjmudd:
Related: https://bugs.mysql.com/bug.php?id=89583. I can't figure out how to build the rpms from the git tree to test this process properly (at least for 8.0.4) as there appear to be no instructions on how this process works.

PR #204, opened 2018-04-22
NDB: Disallow connect string to start with '/'

dveeden:
One can easily use -c to specify the config file by mistake.
-c is connect-string. -f is config file.

Confusing these is bad enough, but ndb_mgmd often accepts a
bogus value for the -c option (e.g -c /etc/ndb/config.ini)

This commit only works if the -c option gets a full path, it won't work for relative paths. Feel free to throw my commit away and implement a different fix for this issue.


dveeden:

PR #208, opened 2018-05-25
Bug #79487: Provide non-Windows implementations for LF_BACKOFF

akopytov:
This is a 8.0.11-based version of a previously contributed patch.


akopytov:

PR #210, opened 2018-05-27
Bug #81814: InnoDB adaptive hash index uses a bad partitioning algorithm

akopytov:
This is a 8.0.11-based version of a previously contributed patch.


akopytov:

PR #214, opened 2018-06-27
Catch exception by reference instead of by value

foobar:
It doesn't make sense to make a copy of the exception


foobar:

PR #217, opened 2018-08-05
 Configurable compression level for archive storage engine

gc-code:
I have added a system variable that allows the user to configure the compression level used by the archive storage engine. This was listed as one of Brian's TODOs and I think it would make a nice addition to the engine. Find below a comparison of the ARZ file sizes for different values of 'archive_compression_level' generated by the windows 'dir' command:

```
05/08/2018  12:04         4,259,928 10_employees.ARZ     :: Compression level 10 (max)
05/08/2018  12:18         4,816,984 0_employees.ARZ      :: Zlib default compression
05/08/2018  12:12         9,076,827 1_employees.ARZ      :: No compression
```

The prefixed number is the compression level. I have shifted Zlib's default compression levels up by one because the variable list doesn't support signed integers. For example, the default compression level is now 0 instead of -1.

I intend on continuing to work on TODOs around the code if you like what I have done here. Constructive feedback is appreciated.

Thanks,
George

### DIFF
--- a/storage/archive/azio.cc
+++ b/storage/archive/azio.cc
@@ -64,7 +64,7 @@ extern "C" PSI_file_key arch_key_file_data;
 */
 int az_open(azio_stream *s, const char *path, int Flags, File fd) {
   int err;
-  int level = Z_DEFAULT_COMPRESSION; /* compression level */
+  int level = srv_compression_level - 1; /* compression level */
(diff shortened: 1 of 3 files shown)


gc-code:
OCA submitted

gc-code:

PR #218, opened 2018-08-12
Fixed build with WolfSSL

gc-code:
WolfSSL does not support X509_check_ip_asc as of version 3.14.0, resulting in build errors. I have added a separate conditional to handle WolfSSL that fixes the issue.

TO REPRODUCE: Attempt to build the 8.0 source with WolfSSL 3.14.0.


gc-code:

PR #220, opened 2018-08-17
Bug #92046: mysql client crashes in libedit/terminal.c

bell-sw:
The fix prevents building incorrect configuration without implicit API declaration.
sys.h and terminal.c sources cleanup:
  termcap library #include and special case definitions moved to single place in terminal.c
  outdated comments and definitions removed
  copyright: year updated


bell-sw:
Let me close this pull request. The change will be submitted from another github user.

PR #221, opened 2018-09-03
Bug #92046: mysql client crashes in libedit/terminal.c

bulasevich:
There only idea of the fix is to raise compile time error when term/curses API is not defined properly. The API functions should be defined either explicitly, or with proper #include directive - otherwise on 64-bit systems we get crash (because C compiler casts 64-bit return value pointer to 32-bit int).

It was a long story about this API. Very similar crash issue was already met and fixed for tgoto function. The API was included explicitly several times - under #if defined(__sun), and under #ifdef notdef, and under #if !HAVE_DECL_TGOTO directives. There exist outdated <termcap.h> api include directive - disabled by #if 0. And, finally, <term.h> and <curses.h> is included. It is a time to clean it up a bit. Let us remove outdated declarations, and put all the term/curses stuff to terminal.c file.

### DIFF
--- a/extra/libedit/sys.h
+++ b/extra/libedit/sys.h
@@ -1,7 +1,7 @@
 /*	$NetBSD: sys.h,v 1.17 2011/09/28 14:08:04 christos Exp $	*/
 
 /*-
- * Copyright (c) 1992, 1993
+ * Copyright (c) 1992, 2018
  *	The Regents of the University of California.  All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


bulasevich:
The OCA has been signed on behalf of BellSoft LLC and the account are activated now in issue tracker and GitHub.

bell-sw:

bulasevich:

PR #224, opened 2018-10-13
BUG#71386 STR_TO_DATE in TRADITIONAL sql_mode fails to parse time-only field

grooverdan:
Problem:

Time only parsing in STR_TO_DATE obeys NON_ZERO_DATES in sql_mode strictly such
that time only strings fail.

Fix:

Only check the NON_ZERO_DATE if the resulting time is DATETIME or DATE.


grooverdan:

PR #228, opened 2018-11-09
my_getncpus based on threads available

grooverdan:
Detecting the cpus based on sysconf of the online CPUs can significantly
over estimate the number of cpus available.

Wheither via numactl, cgroups, taskset, systemd constraints, docker
containers and probably other mechanisms, the number of threads mysqld
can be run on can be quite less.


grooverdan:

PR #233, opened 2018-12-06
Bug#93240 wait_timeout error not clear and not sent to client.

mjonss:
The log entry for a client aborted due to wait_timeout exceeded
is not clearly stated. So this creates a specific error for it.

Also from the client perspective it only sees that the server closed
the connection, not why, so the error will be sent back to
the client before closing the connection.

Notes:
This patch does not work with YaSSL, since it does not clear its
errors in report_errors() in debug build, and ERR_clear_error is
not supported in release build. This results in write will fail
on the previous non-cleared read error (from wait_timeout).

Also debug build non-patched clients will fail an assertion,
due to it only expects async returned result/error in case of COM_QUIT.

And adding protection in net_read_packet like:
  /* Socket can't be used for reading */
  if (net->error == 2 || net->error == 4)
    return packet_error;
will make some rpl tests fail.

### DIFF
--- a/include/mysql_com.h
+++ b/include/mysql_com.h
@@ -359,6 +359,13 @@ typedef struct st_net {
   */
   unsigned char *unused;
   unsigned int last_errno;
+  /**
+    1 - Error, but socket still usable
+    2 - Error, do not use the socket
+    3 - Not used, but will result in a recoverable error.
+    4 - Read error, OK to try write the error and then close the socket.
+    5 - Write error, OK to try read the error and then close the socket.
+  */
   unsigned char error; 
   my_bool unused4; /* Please remove with the next incompatible ABI change. */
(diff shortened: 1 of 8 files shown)


mjonss:
> Hi, thank you for submitting this pull request. In order to consider your code we need you to sign the Oracle Contribution Agreement (OCA). Please review the details and follow the instructions at http://www.oracle.com/technetwork/community/oca-486395.html
> Thanks

I am included in the booking.com OCA, also states '(OCA)' after my name (as reporter) in the bug: https://bugs.mysql.com/bug.php?id=93240

mysql-admin:
Hi Mattias,
Please ignore above extra bot messages - contribution has been accepted.
Your github user was missing from the internal record so the bot didn't recognize you

Thanks for the contribution
==Omer

PR #234, opened 2018-12-17
Impossible WHERE for a!=a, a<a, a>a

grooverdan:
For a table column `a`, the above expressions logically
equate to false in all cases.

With this patch the optimizer knows about this and queries
like:

SELECT * FROM t1 WHERE a!=a

no longer need to evaluate a!=a for every row.

The same applies if the expression was `a<a`, or `a>a`

An `EXPLAIN SELECT COUNT(*) FROM t1 WHERE a<a` will show:

EXPLAIN SELECT COUNT(*) FROM t1 WHERE a<a;
id     select_type     table   partitions      type    possible_keys   key     key_len ref     rows    filtered        Extra
1      SIMPLE  NULL    NULL    NULL    NULL    NULL    NULL    NULL    NULL    NULL    Impossible WHERE

Similarly `NOT (a!=a)` is always true.

EXPLAIN SELECT COUNT(*) FROM t1 WHERE not (a!=a);
id     select_type     table   partitions      type    possible_keys   key     key_len ref     rows    filtered        Extra
1      SIMPLE  NULL    NULL    NULL    NULL    NULL    NULL    NULL    NULL    NULL    Select tables optimized away
Warnings:
Note   1003    /* select#1 */ select count(0) AS `COUNT(*)` from `test`.`t1` where 1


grooverdan:

PR #235, opened 2018-12-28
Add new plugin type: sql_shim to the 8.0 branch

greenlion:
Normal MySQL rewrite plugins can't run queries against the database, as there is no interface for running queries and getting results in the server.  This PR adds such an interface, along with a shim in do_command() to run a plugin to (optionally) rewrite queries.  The provided example plugin includes the SHOW PASSWORD command, which will look up the authentication_string in the mysql.user table and display it to the user.  This type of functionality is impossible to do with rewrite plugins.


greenlion:
There were unintended deletions in this PR.  Closing and will submit a cleaner version.

PR #236, opened 2018-12-28
New SQL shim plugin type for better query rewriting

greenlion:
Normal MySQL rewrite plugins can't run queries against the database, as there is no framework for running queries and getting results in the server.  This PR adds such an interface, along with a shim in do_command() to run a plugin to (optionally) rewrite queries.  The provided example plugin includes the SHOW PASSWORD command, which will look up the authentication_string in the mysql.user table and display it to the user.  This type of functionality is impossible to do with rewrite plugins. 

I was asked why I didn't use Component Services, and wasn't familiar with them.  From what I understand component services aren't meant for the critical path, and do_command() is definitely part of that path.


icecoffee92:
Is this complete or does it need more work?

PR #237, opened 2018-12-30
Add log_bin_implicit_delete setting

dveeden:
On a setup with a master, two intermediate masters and a slave like this:
```
           A
         /   \
        B     C
       /
      D
```

If you have a table like this:
```
CREATE TABLE t1 (id INT PRIMARY KEY) ENGINE=MEMORY;
INSERT INTO t1 VALUES (1);
```

And then you restart C and do a `SELECT * FROM t1`:
* the MEMORY engine marks the table as implicitly emptied
* the server writes a `DELETE FROM test.t1` to the binlog
* If GTID is enabled this creates a GTID local to C

If you now try to move D to C this happens:
* If the DELETE is still in the binary log it will fetch it and execute it, this is OK.
* If the DELETE has been purged from the binlog (e.g. after `binlog_expire_logs_seconds`)
  then replication will fail with `ER_MASTER_HAS_PURGED_REQUIRED_GTIDS`

This commit does two things:
1. It adds a comment to the DELETE to make it easier to discover where it came from.
2. It adds a `log_bin_implicit_delete` to allow the administrator to configure the
   server not to write these DELETE's to the binlog.

* https://dev.mysql.com/doc/refman/8.0/en/replication-features-memory.html
* https://dev.mysql.com/doc/refman/8.0/en/replication-gtids-auto-positioning.html
* https://bugs.mysql.com/bug.php?id=77729

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -1085,6 +1085,7 @@ ulong binlog_stmt_cache_use = 0, binlog_stmt_cache_disk_use = 0;
 ulong max_connections, max_connect_errors;
 ulong rpl_stop_slave_timeout = LONG_TIMEOUT;
 bool log_bin_use_v1_row_events = 0;
+bool log_bin_implicit_delete = 1;
 bool thread_cache_size_specified = false;
 bool host_cache_size_specified = false;
 bool table_definition_cache_specified = false;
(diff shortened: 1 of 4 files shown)


dveeden:

PR #240, opened 2019-01-14
Add system_user as connection attribute in mysql client

dveeden:
This is similar to SYSTEM_USER in the syslog implementation in
mysql client.


dveeden:
This makes it easier to notify people about long running adhoc queries when they are using an application account or some role/proxy account.

dveeden:

PR #241, opened 2019-01-15
THR_LOCK_open reduce usage

grooverdan:
By considering `my_*_opened` as statistical counters and file descriptors as unique with a little restructuring we can eliminate lots of THR_LOCK_open usage.


grooverdan:

PR #242, opened 2019-01-15
my_fclose - retry on close fail like my_close

grooverdan:
my_close will retry closing for non-Windows platforms.

As fclose exhibits the same return behavior, make it behave
like my_close.


grooverdan:

PR #244, opened 2019-01-15
Linux use O_TMPFILE for create_temp_file

grooverdan:
O_TMPFILE creates a tempfile not attached to a filename.

We preserve a state O_TMPFILE_works because kernel version or filesystem could cause failure, most likely permanently.

Documentation needing update after merge - https://dev.mysql.com/doc/refman/8.0/en/internal-temporary-tables.html#internal-temporary-tables-engines "Temporary files are deleted immediately after they are created and opened"


grooverdan:

PR #247, opened 2019-01-25
Updated mysqld_multi script to be able to pass custom parameters to mysql

acampoh:
This also allows to override values already defined in config if
needed without needing to touch the original config files.

### DIFF
--- a/scripts/mysqld_multi.sh
+++ b/scripts/mysqld_multi.sh
@@ -22,7 +22,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA
 
-use Getopt::Long;
+use Getopt::Long qw(:config pass_through);
 use POSIX qw(strftime getcwd);
 use File::Path qw(mkpath);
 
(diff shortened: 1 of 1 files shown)


emottola:
This is useful when a multi-instance MySQL server is used with async replication and/or with Galera libraries. For example to start a node without replication enabled, or bootstrap a new Galera cluster or to perform some special one-time operation like the _wsrep-recovery_.

Without this new feature we need to use a dedicated file per each instance when we need to perform that kind of operation, loosing the ability to manage all the instance with a single my.cnf file.

Starting a mysql async slave without replication enabled:
`mysqld_multi start 1 --skip-slave-start`

Bootstrapping a new Galera cluster can be done with this change very easily:
`mysqld_multi start 1 --wsrep-new-cluster --wsrep_cluster_address='gcomm://'`

In this case the parameters give by command line will override the ones in the _my.cnf_ file, giving us the ability to not touch a working config file.

Perform a recovery of a desynced node become:
`mysqld_multi start 1 --wsrep-recovery`

We had the need to perform this kind of operations, sometimes during outages, so creating each time a dedicated config file that particular instance is not suitable, and as far I can see is the only solution at the moment.

We're already using this patch internally on all our MySQL multi-instance boxes, keeping one single _my.cnf_ configuration file easy to read and maintain.

acampoh:
Done!

acampoh:

emottola:

PR #252, opened 2019-03-03
Call processor in visit_item for FUNC_ITEM

dveeden:
The use case:
Be able to use a post-parse plugin that acts on specific functions being called.

This is done by just adding `FUNC_ITEM` to the list.
However multiple comments seem to indicate that visit function is only called for literals.

The solutions I see are:

1. Splitting this into two processor functions, one for literals and one for functions. However `parse_node_visit_function` etc. have generic names, so to make this clean these then have to be renamed to something like `pare_node_visit_literal_function` etc.
2. Keep this in one processor, but allow one to get the type of the item, this allows one to filter out `FUNC_ITEM` and any other item when needed. That's what is in this pull request.

Please let me know if you think this is a good solution or if there is a better or cleaner solution.

### DIFF
--- a/include/mysql/service_parser.h
+++ b/include/mysql/service_parser.h
@@ -174,6 +174,17 @@ extern struct mysql_parser_service_st {
 
 
   /**
+    Returns the MYSQL_ITEM type as integer
+
+    @param item The item to get the type of.
+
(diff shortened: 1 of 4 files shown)


dveeden:

PR #253, opened 2019-03-04
LOCK INSTANCE FOR BACKUP status

altmannmarcelo:
Currently there is no easy way to know if the instance
is under backup lock. Purge of binary logs for example,
won't happen while backup lock is ongoing.

This commit adds a new status variable named
Ongoing_instance_backup_lock, it displays the integer ENUM
defined by Is_instance_backup_locked_result.


geoyws:
there's a typo in your branch name... it should be named `altmannmarcelo:locked_instance_status` <-- an `n` is missing from `instance` (just pointing it out for a neater Git history)

altmannmarcelo:

PR #259, opened 2019-03-27
fixed broken GCS error message

samitani:
Fixed broken error message of MySQL Group Replication.

When empty `group_replication_local_address` specified, the error message is truncated like below.
```
[GCS] Invalid hostname or IP address (
```
Or with wrong `group_replication_local_address` like `172.17.220.AAA`, it shows only the first char of specified value.
```
[GCS] Invalid hostname or IP address (1) assigned to the parameter local_node!
```

This is expected message.
```
[GCS] Invalid hostname or IP address (172.17.220.AAA:1111) assigned to the parameter local_node!
```


samitani:
I've signed OCA. Let me PR again.

PR #263, opened 2019-05-03
bug #95233: check constraints now support case/if functions

grooverdan:


grooverdan:

PR #266, opened 2019-05-19
InnoDB Data File Purge Asynchronously

xpchild07:
Summary:
    --------
    InnoDB big data file unlink will cause serious stable problem on POSIX file system,
    So InnoDB launch a background thread to truncate data file little by little.

    The data file will firstly renamed into a temporary file when single tablespace is deleted,
    then File Purge thread will truncate file asynchronously and slowly.

    Add new DDL log (PURGE_FILE_LOG) to promise the DDL atomic.

    Several variables control the behaviors:
      INNODB_DATA_FILE_PURGE: Whether enable the async purge strategy
      INNODB_DATA_FILE_PURGE_IMMEDIATE: Unlink data file rather than truncate
      INNODB_DATA_FILE_PURGE_ALL_AT_SHUTDOWN: Cleanup all when normal shutdown
      INNODB_DATA_FILE_PURGE_DIR: Temporary file directory
      INNODB_DATA_FILE_PURGE_INTERVAL: Purge time interval (by milliseconds)
      INNODB_DATA_FILE_PURGE_MAX_SIZE: Purge max size every time (by MB)
      INNODB_PRINT_DATA_FILE_PURGE_PROCESS: Print the process of file purge worker


xpchild07:

PR #267, opened 2019-05-26
Calculating columns sizes based on the character set in mysql client

yurriy:
Currently, print_table_data calculates the size of the column as a length of its longest value **in bytes** (or of its name, if it is longer).
This results in output with additional spaces for multibyte strings. (for utf8, a column can be 2-4 times larger than its longest value).
In this patch column size is calculated based on the character set which client uses.

How to reproduce:
select 'ⲫ' as x;
Output:
```
+-----+
| x   |
+-----+
| ⲫ   |
+-----+
```

Expected output:
```
+---+
| x |
+---+
| ⲫ |
+---+
```


yurriy:

PR #271, opened 2019-06-24
Fixed a mrec_buf_t[] index bug in row_merge_blocks().

Shengliang:
typedef byte        mrec_buf_t[UNIV_PAGE_SIZE_MAX];
sizeof(mrec_buf_t) = sizeof_of(byte)* UNIV_PAGE_SIZE_MAX

buf is a mrec_buf_t[3] type, not a byte[] type.
The index of buf should be 1, not srv_sort_buf_size.


Shengliang:
I need check the same fix to 5.6 and 5.7 branches as well.

PR #274, opened 2019-06-25
BUG#95895 - Shutdown takes long time when table uses AUTOINC and FORE…

satya-bodapati:
…IGN KEY constraints

Problem:
--------
On a database with 2M tables and with buffer pool size of 20GB or more,
shutdown takes approximately 12mins.

Lot of tables should have auto-inc mutexes.

Analysis:
---------
Perf data collected during shutdown revealead about 82% of time is spent in
mutex_destroy().

With a debug instrumentation patch, it is observed that the mutexes destroyed
are AUTOINC.

Still if all tables with AUTOINC mutex are accessed, there would be cached table
objects (dict_table_t). With good amount of 'table_definition_cache',
'table_open_cache_instances' and 'open_files_limit', many tables could remain
in the cache.

The important observation here is about 1.6Million tables in cache and
Count one mutex per each of those.

Still why ~ 10mins (82% of time) for destruction of these 2.6 Million mutexes?

These are custom mutexes. Implemented by InnoDB and not pthread_mutexes.

For these mutexes, to collect statistics, a Counter is created for each mutex.
Counter has information about spins, waits, calls. One counter is created for
every mutex instance. So for AUTOINC Mutex type, there is one vector which
will hold all Counters.

For AUTOINC, the vector would hold 1.6Million Count elements

During shutdown, each mutex is destroyed and so that removes the stat object(count)
from the huge vector. This is one-by-one removal.

Lets see the complexity of removal of an element from vector of size N. To remove
one element, it is O(N). And to remove N elements, it is ~ O(N^2). Lets calculate
this for 1.6Million. So it would be O(1.6Million ^ 2) = 2.5Trillion operations.

Now we know why it took 82% of shutdown time :)

Fix:
----
Similar problem doesn't exist for Block Mutexes because for stats, it uses aggregrated
counters. See BlockMutexPolicy.

For AUTOINC too, which can be in Millions, it doesn't make sense to collect mutex
stats individually. So use the same strategy as BlockMutex. i.e to use Aggregate Stats

Use ib_bpmutex_t instead of ib_mutex_t for Autoinc mutex of table.

### DIFF
--- a/storage/innobase/dict/dict0dict.cc
+++ b/storage/innobase/dict/dict0dict.cc
@@ -750,7 +750,7 @@ dict_table_autoinc_alloc(
 	void*	table_void)
 {
 	dict_table_t*	table = static_cast<dict_table_t*>(table_void);
-	table->autoinc_mutex = UT_NEW_NOKEY(ib_mutex_t());
+	table->autoinc_mutex = UT_NEW_NOKEY(AutoIncMutex());
 	ut_a(table->autoinc_mutex != NULL);
 	mutex_create(LATCH_ID_AUTOINC, table->autoinc_mutex);
 }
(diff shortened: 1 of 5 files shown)


satya-bodapati:

satya-bodapati:
See https://jira.percona.com/browse/PS-5639 for more details

PR #278, opened 2019-07-17
Fix bug https://bugs.mysql.com/bug.php?id=95927

peresypkin:
The PR fixes the bug https://bugs.mysql.com/bug.php?id=95927


peresypkin:

PR #286, opened 2019-12-24
bug#97828 If ANALYZE TABLE runs more than 600 seconds diagnostic query may crash server

kamil-holubicki:
bug#97828

Locking of table statistics limited to the time when statistics related table members are changed. While calculation is done, statistics are not locked, so can be queried.

Index analysis protected by new mutex as it cannot be done concurrently, because it shares the same index object which is modified during analysis.


kamil-holubicki:

PR #289, opened 2020-01-19
Include port number in client error message

dveeden:
With an `~/.my.cnf` like this:
```
[client]
port=1234
```

Before:
```
$ mysql -h 127.0.0.1
ERROR 2003 (HY000): Can't connect to MySQL server on '127.0.0.1' (111)
```

After:
```
mysql -h 127.0.0.1
ERROR 2003 (HY000): Can't connect to MySQL server on '127.0.0.1:1234' (111)
```

This also works if `-P` is used to supply a port number.

This makes troubleshooting connection issues a tiny bit easier.


dveeden:

PR #294, opened 2020-03-11
Set non-zero error code for compression failures

dveeden:
When running `mysqlpump --output-compression foobar` it tries to use foobar as the compression method which will obviously not work.

```
mysqlpump: [ERROR] (0) Unknown compression method: foobar
Dump process encountered error and will not continue.
```
While it says it won't continue it actually does continue.

When the process ends the returncode is taken from the error. However as the error code is 0 the return code is also 0. The result is that the return code doesn't indicate failure.


dveeden:

PR #295, opened 2020-03-12
lz4_decompress, zlib_decompress: Allow these to work as standard filters

dveeden:
* If no output file is specified, then direct output to stdout.
* If no input file is specified, then read from stdin.

This allows one to use pipes and standard UNIX tools to work with zlib
and lz4 compressed SQL dumps.


dveeden:

PR #300, opened 2020-06-19
Bug#99935 : innodb_doublewrite_files is not correct when innodb_buffe…

satya-bodapati:
…r_pool_size > 1G

Problem:
-------
According to documentation, default value of innodb_doublewrite_files is
innodb_buffer_pool_instances * 2

But when you start InnoDB with buffer pool size > 1G (the number of instances
is 8), it still creates 2 files only. Expected files is 16 (8*2)

Fix:
---
Use correct number of dblwr according to the number of buf pool instances

### DIFF
--- a/storage/innobase/buf/buf0dblwr.cc
+++ b/storage/innobase/buf/buf0dblwr.cc
@@ -1721,7 +1721,9 @@ dberr_t dblwr::open(bool create_new_db) noexcept {
   uint32_t segments_per_file{};
 
   if (dblwr::n_files == 0) {
-    dblwr::n_files = 2;
+    dblwr::n_files = std::max(2UL, srv_buf_pool_instances * 2);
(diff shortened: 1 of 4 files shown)


storage/innobase/buf/buf0dblwr.cc
```diff
@@ -1721,7 +1721,9 @@ dberr_t dblwr::open(bool create_new_db) noexcept {
   uint32_t segments_per_file{};
 
   if (dblwr::n_files == 0) {
-    dblwr::n_files = 2;
+    dblwr::n_files = std::max(2UL, srv_buf_pool_instances * 2);
+  } else if (dblwr::n_files > (srv_buf_pool_instances * 2)) {
```
satya-bodapati:
```suggestion
  } else if (dblwr::n_files > srv_buf_pool_instances * 2) {
```

satya-bodapati:

PR #301, opened 2020-07-06
Bug#100118 Server doesn't restart because of too many gaps in the mysql.gtid_executed table

venkatesh-prasad-v:
Problem & Analysis
------------------
Since the introduction of a dedicated thread for persisting GTIDs of
InnoDB transactions by WL#9211, new behavior for updating the
mysql.gtid_executed table is as follows,

- If binary log is enabled, the mysql.gtid_executed table is updated on
  next binlog rotation (either by FLUSH LOGS or server restart).
- If binary log is disabled or log_slave_updates is disabled (for slave
  threads), then
  - If it is an InnoDB transaction, it leaves it to the gtid persister
    thread to update.
  - Otherwise, it writes its GTID into mysql.gtid_executed table.

On a binlogless slave having both transactional and non-transactional
workloads, as per the current design, when the GTID persister thread
reaches the threshold (either once per 1k transactions or every 1
second), it flushes its list (updates the table) to the
mysql.gtid_executed table and tries to compress the gtid_executed table.
It succeeds compressing only the first few rows (because it filled the
gaps present in the beginning of the table by merging transactional
updates) and fails to compress further rows of the table (because of the
gaps introduced by non-transactional updates).

In every attempt it tries to merge the table, it merges only a few
consecutive rows and leaves the other rows as is, thereby taking more
time for scanning the full table. By the time it finishes the table
scan, the slave applier threads would insert few more entries to the end
of the table. As a result, when there is a high load on the server, if
the GTID persister thread once starts scanning the table, it is more
likely that the scan never stops and thus causes the flush list to grow
causing more gaps in the mysql.executed table.

In addition to the above, on a binlogless slave, explicit gtid
compression requests raised from the below queries also caused the
persister thread to be stuck while compressing the gtid_executed table.

1. FLUSH ENGINE LOGS
2. RESET MASTER, RESET SLAVE and CHANGE MASTER
3. Binary/relay log rotation
4. expire logs_by_seconds and expire_logs_by_days

Fix
---
Made the GTID persister thread to compress the mysql.gtid_executed table
only on explicit request from innobase_flush_logs() when both binary
logging and log_slave_updates are enabled.

As per the current fix, the final truth table of server behavior is:
```
+-------------+----------+-------------------+-------------------+--------------+
|  Is server  |          |                   |  Explicit request |    Result    |
| Standalone/ |  Binlog  | log_slave_updates |    created on     |              |
|    Slave?   | enabled? |      enabled?     | FLUSH ENGINE LOGS |  Persister/  |
|             |          |                   |                   |  Compressor  |
+=============+==========+===================+===================+==============+
|  Standalone |    Yes   |        Yes        |        Yes        |   Persister  |
+-------------+----------+-------------------+-------------------+--------------+
|  Standalone |    Yes   |         No        |         No        |  Compressor  |
+-------------+----------+-------------------+-------------------+--------------+
|  Standalone |    No    |        Yes        |        N/A        |              |
+-------------+----------+-------------------+-------------------+ Invalid case |
|  Standalone |    No    |         No        |        N/A        |              |
+-------------+----------+-------------------+-------------------+--------------+
|    Slave    |    Yes   |        Yes        |        Yes        |   Persister  |
+-------------+----------+-------------------+-------------------+--------------+
|    Slave    |    Yes   |         No        |         No        |  Compressor  |
+-------------+----------+-------------------+-------------------+--------------+
|    Slave    |    No    |        Yes        |        N/A        | Invalid case |
+-------------+----------+-------------------+-------------------+--------------+
|    Slave    |    No    |         No        |         No        |  Compressor  |
+-------------+----------+-------------------+-------------------+--------------+
```

### DIFF
--- a/mysql-test/suite/binlog_gtid/r/binlog_gtid_compressor_validation.result
+++ b/mysql-test/suite/binlog_gtid/r/binlog_gtid_compressor_validation.result
@@ -1,0 +1,28 @@
+RESET MASTER;
+#
+# 1. Create an InnoDB table and insert 2 rows.
+#    Wait till each entry appears individually in the table.
+CREATE TABLE t1 (a INT NOT NULL PRIMARY KEY) ENGINE = INNODB;
+INSERT INTO t1 VALUES (1);
+INSERT INTO t1 VALUES (2);
+#
+# 2. Assert that table is not compressed.
+include/assert.inc [COMMITTED_GTIDS SERVER_UUID:1-3]
+include/assert.inc [The mysql gtid table should have 3 rows.]
+#
+# 3. Enable debug points to enable force compression and to print the
+#    compression info for validation purpose.
+# Adding debug point 'simulate_force_compress' to @@GLOBAL.debug
+# Adding debug point 'print_gtid_compression_info' to @@GLOBAL.debug
+#
+# 4. Insert third row and assert that table is compressed.
+INSERT INTO t1 VALUES (3);
+include/assert.inc [The mysql gtid table should have 1 row.]
+
+# 5. Assert that the compression was done by the gtid compressor thread.
+include/assert_grep.inc [GTID table compression is done by compressor thread.]
+#
+# 6. Cleanup
+# Removing debug point 'simulate_force_compress' from @@GLOBAL.debug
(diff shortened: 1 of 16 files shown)


venkatesh-prasad-v:

PR #303, opened 2020-07-27
bug100335: The udf_init() function cause server crash.

zbdba:


zbdba:

PR #305, opened 2020-08-25
Aarch64 support

tsahee:
After signing the OCA was sorted out, this branch fixes two items missing in aarch64:
* an implementation of UT_RELAX_CPU macro
* enabling runtime conditional LSE atomics support, using the -moutline-atomics gcc compiler flag


tsahee:

PR #314, opened 2020-11-22
Use mmap for binlog

zhjwpku:
The main point of this patch is to use mmap to accelerate binlog
writing speed, it's fairly an easy task to implement due to the
inherant hirarchy. I just need to add a new class that extend
`Truncatable_ostream`(see `MMAP_ostream` for detail) and maintain
the underling mmap info in `struct MMAP_INFO`.

But there is something that need to point out.

1. The mmaped file will be truncated to the max_binlog_size at
   the beginning, so the real events size will not be showed
   when `ls` the binlog files.

2. When some command is reading the current writting binlog file,
   for example `show binlog events in '<cur_writting_binlog>'`
   uses Binlog_event_data_istream::read_event_header to read the
   file, as this file was truncated to the max_binlog_size, it
   will continue to read '0's after the last event in that file.
   I add a special condition to solve this problem, that is

   if (m_event_length == 0 && m_cur_binlog)
    return m_error->set_type(Binlog_read_error::READ_EOF);

3. relay log hasn't been tested

4. during recover process, use IO_CACHE_ostream to handle existed
   binlog

Signed-off-by: Zhao Junwang <zhjwpku@gmail.com>

### DIFF
--- a/include/my_sys.h
+++ b/include/my_sys.h
@@ -541,6 +541,35 @@ inline size_t my_b_bytes_in_cache(const IO_CACHE *info) {
   return *info->current_end - *info->current_pos;
 }
 
+/* This struct is used for mmap for binlog */
+struct MMAP_INFO {
+  // end position of file
+  my_off_t end_pos_of_file{0};
+  // mmap return address
+  uchar *addr{nullptr};
+  // next write position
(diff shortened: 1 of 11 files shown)


zhjwpku:

PR #318, opened 2020-12-31
cloning: Wait for 15m instead of 5m before dropping snapshot

dveeden:
5m is too short to allow resumption in some cases where the network has
a short moment of instability. Increasing this to 15m isn't adding too
much overhead and could avoid having to copy multiple terabytes again.


dveeden:

dveeden:

PR #322, opened 2021-01-29
correct wrong debug assert in cmp_rec_rec_simple

zhjwpku:
The debug assertion wil be no effect since two params are exactly the same.
This must be a typo.


zhjwpku:

PR #324, opened 2021-02-10
fix IB_VEC_OFFSET 's wrong parameter

zhjwpku:
This is not a bug since the four places using this MACRO
has a `vec` local variable in the calling scope, but this may
be some problem when there is a `vec` vector and a `vec2`
vector, because IB_VEC_OFFSET(vec2, i) will give the offset
of `vec`, thus I think this is a potential bug that we should
address.


zhjwpku:

PR #325, opened 2021-02-19
Bug #102647: Replica sql_thread getting stuck due to MASTER_DELAY

venkatesh-prasad-v:
Problem
-------
The replication applier thread in a multi-threaded replica randomly gets
stuck in

  Waiting until MASTER_DELAY seconds after master executed event

and doesn't apply any transaction for a long time when the replica is
configured with MASTER_DELAY.

Background
----------
After the bugfix of BUG#17326020 (commit: 3f6ed37f), when the receiver
thread reconnects to its source using GTID auto positioning while in the
middle of a transaction, it leaves the partial transaction on the
relaylog as is and will fully retrieve the same transaction again. While
doing so, the receiver thread, as per GTID auto-position protocol will
write master's FORMAT_DESCRIPTION_EVENT with log_pos != 0 into the relay
log after each (re)connection.

This FORMAT_DESCRIPTION_EVENT shall look like below in the relay log.
```
  ===BEGIN===
  # at 4
  #210215 16:18:05 server id 1  end_log_pos 123 CRC32 0x1120fc25 	Start: binlog v 4, server v 5.7.32-35-debug-log created 210215 16:18:05
  # This Format_description_event appears in a relay log and was generated by the slave thread.
  # at 123
  #210215 16:18:05 server id 1  end_log_pos 154 CRC32 0xa29c149c 	Previous-GTIDs
  # [empty]
  # at 154
  #700101  5:30:00 server id 1111  end_log_pos 0 CRC32 0x93c56c74 	Rotate to mysql-bin.000001  pos: 4
  # at 201
  #210215 14:19:18 server id 1111  end_log_pos 123 CRC32 0x25802c3f 	Start: binlog v 4, server v 5.7.32-35-debug-log created 210215 14:19:18 at startup
  ROLLBACK/*!*/;
  BINLOG '
  jjUqYA9XBAAAdwAAAHsAAAAAAAQANS43LjMyLTM1LWRlYnVnLWxvZwAAAAAAAAAAAAAAAAAAAAAA
  AAAAAAAAAAAAAAAAAACONSpgEzgNAAgAEgAEBAQEEgAAXwAEGggAAAAICAgCAAAACgoKKioAEjQA
  AT8sgCU=
  '/*!*/;
  # at 320
  #210215 16:18:05 server id 0  end_log_pos 367 CRC32 0xdd12fd05 	Rotate to mysql-bin.000001  pos: 154
  # at 367
  ===END===
```

When the applier thread starts applying the FORMAT_DESCRIPTION_EVENT (at
position 201), it checks if it is in the middle of a group. If it is, it
will queue to the current worker a QUERY(ROLLBACK) event to make the
worker gracefully finish its work before starting to apply the new
(complete) copy of the group, and only after that it will let the MTS
coordinator to apply the ROTATE_LOG_EVENT (at position 320) in
synchronous mode.

Analysis
--------
In our tests, it was observed that the replication applier thread got
stuck when it was applying the Query_log_event (ROLLBACK) as part of
handling the partial relay log event. On further investigation, it was
found that the Query_log_event (for BEGIN/ROLLBACK) created by the
applier thread had the current time as its timestamp. As a result, the
delay function `slave_sleep()` slept for MASTER_DELAY seconds since the
event timestamp was less than the evaluated current timestamp.

Fix
---
Make the BEGIN/ROLLBACK query to execute with same timestamp as that of
the master's FORMAT_DESCRIPTION_EVENT in
`coord_handle_partial_binlogged_transaction()` function.

This patch additionally also adds a .cnf file for the test
`rpl_mts_gtids_restart_slave_io_lost_trx.test` so that the test is not
skipped on MTR runs with default options.

### DIFF
--- a/mysql-test/extra/rpl_tests/rpl_gtids_restart_slave_io_lost_trx.test
+++ b/mysql-test/extra/rpl_tests/rpl_gtids_restart_slave_io_lost_trx.test
@@ -10,6 +10,11 @@
 # IO thread right after queuing an event of a given type, starting the IO
 # thread again to fully retrieve the partially received transaction.
 #
+# Additionally, this test also tests partial transactions with MASTER_DELAY
+# configured and verifies that the applier thread doesn't wait for MASTER_DELAY
+# when the receiver thread is restarted in the middle of a transaction, if MTS
+# and GTID autoposition protocol. To test this, set $master_delay_test=1 and
+# source this file.
 
 --source include/have_gtid.inc
 
@@ -33,6 +38,14 @@ CREATE TABLE t1(i INT) ENGINE=InnoDB;
 # their replicated versions in the slave's relaylog).
 FLUSH LOGS;
 --source include/sync_slave_sql_with_master.inc
+
+# Test for MASTER_DELAY with partial transactions.
+if ($master_delay_test)
+{
+  --source include/stop_slave.inc
+  CHANGE MASTER TO MASTER_DELAY=5;
+  --source include/start_slave.inc
+}
 SET @save_debug=@@global.debug;
 
 #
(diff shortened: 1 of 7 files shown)


venkatesh-prasad-v:

PR #328, opened 2021-03-05
Update makelist to support CRLF files

yezhiweicmss:
Doing makelist -h file which with CRLF line terminators went wrong.
Update makelist to support CRLF files.


yezhiweicmss:
i can't receive the verifying email from oracle-acct_ww@oracle.com when i am creating an oracle account.

PR #331, opened 2021-03-23
fix priority in comparison

ihsinme:
in the original version, the assignment will be performed after comparison.


ihsinme:
i signed an agreement

ihsinme:

PR #334, opened 2021-04-12
innodb_dedicated_server = 1 with VM with 2GB of RAM causes startup to fail

adamcable:
If the VM has 2GB of RAM, and you use the innodb_dedicated_server = 1 declaration, then the server fails to boot as it only creates a single log file.
This is because the current code doesn't cater for when the automatic buffer pool size is set to 1GB.
The amended code ensures that at least 2 log files are created.


adamcable:
This is a fix for bug #103291
https://bugs.mysql.com/bug.php?id=103291

adamcable:
> Hi, thank you for submitting this pull request. In order to consider your code we need you to sign the Oracle Contribution Agreement (OCA). Please review the details and follow the instructions at https://oca.opensource.oracle.com/
> Thanks

Thanks, now completed

adamcable:

PR #338, opened 2021-05-22
Add default variable value information to performance_schema.variable…

ejortegau:
This PR addresses the feature request reported under [bug #103779](https://bugs.mysql.com/bug.php?id=103779). Basically, it allows the client to obtain information about the default value of all variables.

It does so by adding column `default_value` to `performance_schema.variables_info`:

```sql
mysql> select variable_name, default_value from variables_info where variable_name = 'autocommit';
+---------------+---------------+
| variable_name | default_value |
+---------------+---------------+
| autocommit    | ON            |
+---------------+---------------+
1 row in set (0.00 sec)

mysql> select variable_name, default_value from variables_info where variable_name = 'innodb_buffer_pool_size';
+-------------------------+---------------+
| variable_name           | default_value |
+-------------------------+---------------+
| innodb_buffer_pool_size | 134217728     |
+-------------------------+---------------+
1 row in set (0.01 sec)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,7 @@ source_downloads
 
 # Configuration files for Visual Studio Code
 .vscode/
+
+# CLion config and build files
+.idea/
+cmake-build-debug/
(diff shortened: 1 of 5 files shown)


ejortegau:
I am withdrawing the PR, as I realized it is not providing the default value of the variable, but its current value. This adds no value, as this information is already available via `perofrmance_schema.global_variables` and `SHOW GLOBAL VARIABLES`.

I may try to fix this in and re-open, or create a new PR, in the future.

PR #345, opened 2021-06-27
Refactoring redundancy code.

16yuki0702:
The '''else''' block is no need, so deleted it with this PR for refactoring.


16yuki0702:

PR #348, opened 2021-07-14
Bug #98495: Timestamp is not set to CURRENT_TIMESTAMP in mysql.tables…

venkatesh-prasad-v:
…_priv

Problem
-------
All the grant tables have a column called Timestamp to display when the
last change occurred for a user.

However, it was not fully implemented and caused the Timestamp field of
the mysql.*_priv table to be set to NULL timestamp("0000-00-00
00:00:00") for all GRANT/REVOKE queries and was preventing the logical
restore of grant tables to zero date.

Fix
---
We now write the the query start time as the timestamp for the
GRANT/REVOKE queries.

NOTE
----
For older datadirs, even after the upgrade, the existing entries can
still have zero timestamps. In such cases, for performing a logical
backup, users are expected to

1. Run GRANT/REVOKE query and then take backup.

   OR

2. Update the timestamp of the mysql.*_priv tables.

   Eg: UPDATE mysql.tables_priv SET Timestamp = CURRENT_TIMESTAMP WHERE User = 'u1';

   OR

3. Take backup and then edit the dump file and replace zero timestamp in the
   with CURRENT_TIMESTAMP.

  For example:

  change the contents
    REPLACE INTO `tables_priv` VALUES ('%','test','u1','t1','root@localhost','0000-00-00 00:00:00','','Select');
  to
    REPLACE INTO `tables_priv` VALUES ('%','test','u1','t1','root@localhost',CURRENT_TIMESTAMP,'','Select');

Testing Done
---
```
Using parallel: 1

==============================================================================
                  TEST NAME                       RESULT  TIME (ms) COMMENT
------------------------------------------------------------------------------
[ 20%] main.hostname_length                      [ pass ]    463
[ 40%] main.mysqlpump_long_hostname              [ pass ]    639
[ 60%] main.timestamps_in_priv_tables            [ pass ]    493
[ 80%] perfschema.hostname_length                [ pass ]    295
------------------------------------------------------------------------------
The servers were restarted 0 times
The servers were reinitialized 0 times
Spent 1.890 of 19 seconds executing testcases

Completed: All 5 tests were successful.
```


venkatesh-prasad-v:

PR #349, opened 2021-07-23
fix fts_optimize_thread missing in performance_schema.threads

YKG:
`fts_optimize_thread` and `buf_resize_thread` are missing in the output of `select * from performance_schema.threads`.

This patch is to fix the first one, I'll submit [another PR](https://github.com/mysql/mysql-server/pull/350) to fix the second one.


YKG:
Hi, I have signed OCA before I create this PR

PR #350, opened 2021-07-23
fix buf_resize_thread missing in performance_schema.threads

YKG:
`fts_optimize_thread` and `buf_resize_thread` are missing in the output of `select * from performance_schema.threads`.

This PR is to fix the second one. The patch for the first is here: https://github.com/mysql/mysql-server/pull/349


YKG:
Hi, I have signed OCA before I create this PR @mysql-oca-bot

PR #352, opened 2021-07-27
Improved the expired password logging sent to the error log

sjmudd:
The previous text provided no information on who the user was or where it was logging in from.

Previous sample logging showed:

```
2021-07-22T10:54:09.718676Z 2671647 [Note] [MY-010928] [Server] Your password has expired. To log in you must change it using a client that supports expired passwords.
```

Modified logging shows:

```
2021-07-27T12:00:32.420060Z 105 [Note] [MY-013776] [Server] Login failed for user 'someuser'@'somehost.example.com' due to an expired password.
```

I'm not sure if anything else is needed. If so please let me know.


sjmudd:

sjmudd:
related: https://bugs.mysql.com/bug.php?id=104428

PR #354, opened 2021-08-03
Add endpoint host:port to error message and describe another failure scenario

sjmudd:
A minor change to improve reporting the endpoint which help diagnosing the error.

See: https://bugs.mysql.com/bug.php?id=104523

I think it would be good to also be able to add `Reason: [os_error_number] os text of the error` to the text but was not sure how to do that.  The basic patch I have provided helps I think. Improving it further would be good too.


sjmudd:

PR #355, opened 2021-08-06
[typo] microseconds --> milliseconds

cighao:
if use "ms" in the output, it should use millisecond for timing (not microsecond)


cighao:
The agreement is now signed.

cighao:

PR #356, opened 2021-08-10
Bug #104575	Duplicated threads in performance_schema.threads table

YKG:
I explained the root cause in the [bug report](https://bugs.mysql.com/bug.php?id=104575).


YKG:

PR #357, opened 2021-08-11
Use system CA roots for client if no option provided

dbussink:
Right now a CA chain is required to be provided in verify_ca or verify_identity mode. This makes it much more cumbersome when writing a client that connects to a server where the certificate in the system roots and when it provides a proper certificate signed by such an authority.

By falling back to the system roots, deployments are greatly simplified because it removes the need to manually specify the root chain. The root chain is highly operating system and deployment dependent, so it requires a different configuration flag on each platform.

By setting up the system roots through OpenSSL, deploying a MySQL server with a root signed installed on a system is greatly simplified for client configuration.

Many clients in other languages / on other platforms already follow a similar behavior where system roots are used if no specific CA chain is configured. This applies to for example Java, Go, Node.js, Rust & .NET.

This change also makes it simpler for drivers that use libmysqlclient such as the ones often used in Ruby and Python to simplify configuration for developers there as well.

### DIFF
--- a/sql-common/client.cc
+++ b/sql-common/client.cc
@@ -4162,20 +4162,6 @@ static int cli_establish_ssl(MYSQL *mysql) {
   }
 
   /*
-    If the ssl_mode is VERIFY_CA or VERIFY_IDENTITY, make sure that the
-    connection doesn't succeed without providing the CA certificate.
-  */
-  if (mysql->options.extension &&
-      mysql->options.extension->ssl_mode > SSL_MODE_REQUIRED &&
-      !(mysql->options.ssl_ca || mysql->options.ssl_capath)) {
-    set_mysql_extended_error(mysql, CR_SSL_CONNECTION_ERROR, unknown_sqlstate,
-                             ER_CLIENT(CR_SSL_CONNECTION_ERROR),
-                             "CA certificate is required if ssl-mode "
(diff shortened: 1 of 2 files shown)


dbussink:
@mysql-oca-bot I have changed jobs since I've signed the OCA and I think it was also under the company name there. This means I might need a new OCA?

mysql-admin:
Hi Dirkjan, I am responsible for MySQL OCAs and will help you. It seems that you will need to have a personal OCA, please go to https://oca.opensource.oracle.com and submit one. Once done, please confirm to me to lenka(dot)kasparova(at)oracle(dot)com and I will process that urgently. Please be aware that you also will need to have a valid account in our bug system (bugs.mysql.com) on your personal email address (you can either modify your current account and try to change the email address, if that is possible..or simply create a new account there with your personal email address). Please let me know if you have further questions,
thank you, lenka

dbussink:
@mysql-admin I suspect that I'll need to get one signed also by my current employer then and not a personal one. Can I start that process on the same site then?

dbussink:
I realized the implementation is wrong here. I'm going to add it to #358 instead because in order to fix this correctly I need to revisit the order of arguments as in that PR as well so they'd immediately conflict with each other if they were separate.

PR #358, opened 2021-08-11
Extend TLS/SSL functionality with SNI and using system roots

dbussink:
This changes enables sending along the SNI extension to the server when connecting over TLS with a client. It's best practice these days to send SNI as a TLS client.

Usage of SNI would also enable uses cases like routing of MySQL connections to different backends behind a single IP.

This change only implements the client side parts. Server side there is also value in providing different certificates on different SNI names, for example for multi homed MySQL servers with different DNS entries to connect combined with verify_identity semantics.

At first though, clients should start sending the extension before it's useful to implement anything on the server side.

Additionally, the second commit here enables usage of the system roots when no CA file or path is configured. Right now a CA chain is required to be provided in verify_ca or verify_identity mode. This makes it much more cumbersome when writing a     client that connects to a server where the certificate in the system roots and when it provides a proper certificate signed by such an authority.

By falling back to the system roots, deployments are greatly simplified because it removes the need to manually specify the root chain. The root chain is highly operating system and deployment dependent, so it requires a different configuration flag on each platform.

By setting up the system roots through OpenSSL, deploying a MySQL server with a root signed installed on a system is greatly simplified for client configuration.

Many clients in other languages / on other platforms already follow a similar behavior where system roots are used if no specific CA chain is configured. This applies to for example Java, Go, Node.js, Rust & .NET.

This change also makes it simpler for drivers that use libmysqlclient such as the ones often used in Ruby and Python to simplify configuration for developers there as well.

### DIFF
--- a/include/violite.h
+++ b/include/violite.h
@@ -265,7 +265,7 @@ struct st_VioSSLFd *new_VioSSLConnectorFd(
     const char *key_file, const char *cert_file, const char *ca_file,
     const char *ca_path, const char *cipher, const char *ciphersuites,
     enum enum_ssl_init_error *error, const char *crl_file, const char *crl_path,
-    const long ssl_ctx_flags, const char *server_host, bool verify_identity);
+    const long ssl_ctx_flags, const char *server_host, int verify, bool verify_identity);
 
 long process_tls_version(const char *tls_version);
 
--- a/include/violite.h
+++ b/include/violite.h
@@ -253,6 +253,7 @@ const char *sslGetErrString(enum enum_ssl_init_error err);
 
 struct st_VioSSLFd {
   SSL_CTX *ssl_context;
+  const char *hostname;
 };
 
 int sslaccept(struct st_VioSSLFd *, MYSQL_VIO, long timeout,
@@ -264,7 +265,7 @@ struct st_VioSSLFd *new_VioSSLConnectorFd(
     const char *key_file, const char *cert_file, const char *ca_file,
     const char *ca_path, const char *cipher, const char *ciphersuites,
     enum enum_ssl_init_error *error, const char *crl_file, const char *crl_path,
-    const long ssl_ctx_flags, const char *server_host);
+    const long ssl_ctx_flags, const char *server_host, bool verify_identity);
 
 long process_tls_version(const char *tls_version);
 
--- a/sql-common/client.cc
+++ b/sql-common/client.cc
@@ -3352,9 +3352,12 @@ const char *STDCALL mysql_get_ssl_cipher(MYSQL *mysql MY_ATTRIBUTE((unused))) {
   return nullptr;
 }
 
+#if OPENSSL_VERSION_NUMBER < 0x10002000L
 /*
   Check the server's (subject) Common Name against the
-  hostname we connected to
+  hostname we connected to. This function is only used
+  when OpenSSL is older than 1.0.2 since versions since
+  1.0.2 provide an implementation for hostname validation.
 
   SYNOPSIS
   ssl_verify_server_cert()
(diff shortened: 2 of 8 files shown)


include/violite.h
```diff
@@ -264,7 +265,7 @@ struct st_VioSSLFd *new_VioSSLConnectorFd(
     const char *key_file, const char *cert_file, const char *ca_file,
     const char *ca_path, const char *cipher, const char *ciphersuites,
     enum enum_ssl_init_error *error, const char *crl_file, const char *crl_path,
-    const long ssl_ctx_flags, const char *server_host);
+    const long ssl_ctx_flags, const char *server_host, bool verify_identity);
```
dbussink: This is split out into a separate argument since before, setting the host implied `verify_identity`. The goal here is though to always send the SNI extension independent of the verify mode, so that's why it's now a separate flag here.

dbussink:
@mysql-oca-bot I have changed jobs since I've signed the OCA and I think it was also under the company name there. This means I might need a new OCA?

sql-common/client.cc
```diff
@@ -4427,14 +4415,19 @@ static net_async_status cli_establish_ssl_nonblocking(MYSQL *mysql, int *res) {
     /* sslconnect creates a new vio, so update it. */
     vio_set_blocking_flag(net->vio, !ctx->non_blocking);
 
+    /* OpenSSL 1.0.2 and up will use their own verification function.
+     * This is set up in new_VioSSLConnectorFd and adding here it here
+     * again is superfluous.
+     */
+#if OPENSSL_VERSION_NUMBER < 0x10002000L
```
dbussink, line 4422: On OpenSSL 1.0.2 and newer, the additional identity check here is not needed since we already set it up inside `new_VioSSLConnectorFd` using the OpenSSL provided validation function. This means the check here was superfluous and would result in double checking the certificate.

If anything, it makes it more brittle because there could be implementation differences between these validation functions (such as honoring DNS SANs, using the deprecated CN or not etc).

vio/viosslfactories.cc
```diff
@@ -795,19 +801,12 @@ struct st_VioSSLFd *new_VioSSLConnectorFd(
...
-    const long ssl_ctx_flags, const char *server_host) {
+    const long ssl_ctx_flags, const char *server_host, int verify, bool verify_identity) {
   struct st_VioSSLFd *ssl_fd;
-  int verify = SSL_VERIFY_PEER;
-
-  /*
-    Turn off verification of servers certificate if both
-    ca_file and ca_path is set to NULL
-  */
-  if (ca_file == nullptr && ca_path == nullptr) verify = SSL_VERIFY_NONE;
```
dbussink, line 806: We no longer infer this value, but it gets passed in explicitly by the caller based on the SSL mode that is selected. This means the SSL mode flag is leading and we fall back already automatically to the system roots.

sql-common/client.cc
```diff
@@ -4161,20 +4152,6 @@ static int cli_establish_ssl(MYSQL *mysql) {
...
-  if (mysql->options.extension &&
-      mysql->options.extension->ssl_mode > SSL_MODE_REQUIRED &&
-      !(mysql->options.ssl_ca || mysql->options.ssl_capath)) {
-    set_mysql_extended_error(mysql, CR_SSL_CONNECTION_ERROR, unknown_sqlstate,
-                             ER_CLIENT(CR_SSL_CONNECTION_ERROR),
-                             "CA certificate is required if ssl-mode "
-                             "is VERIFY_CA or VERIFY_IDENTITY");
-    goto error;
-  }
-
```
dbussink, line 4177: The code itself already supports fallback to the system roots if nothing is given. So this check can be removed on its own. The reason for all the other changes is that providing the ca path or file was also used to infer the SSL verification mode, which is now made all explicit.

The fallback logic exists here:

https://github.com/mysql/mysql-server/blob/beb865a960b9a8a16cf999c323e46c5b0c67f21f/vio/viosslfactories.cc#L696-L700

dbussink:
Uhm, what? Things were still in progress afaik? Can this please be reopened?

mysql-admin:
@dbussink Sorry about that. The bot is not as clever as us humans and was not aware of the change in OCA being worked on. 
It should post a note about waiting for an OCA and will wait until a new one is in place
==Omer

dbussink:

Phanatic:
We're interested in responding to any feedback on this PR and continue to work with mysql community to get this merged, Can this please be reopened?

dbussink:
@Phanatic I think things were moved to https://bugs.mysql.com/bug.php?id=104649 but sadly there's not much activity there yet.

PR #359, opened 2021-08-14
add query-attributes options to mysql client

YKG:
This PR is intended to add an option to mysql client to disable query attributes.
See https://bugs.mysql.com/bug.php?id=104624 for more details.


YKG:

PR #364, opened 2021-08-27
fix buf_resize_thread missing in performance_schema.threads

YKG:
buf_resize_thread are missing in the output of select * from performance_schema.threads.

See: https://bugs.mysql.com/bug.php?id=104583


YKG:

PR #365, opened 2021-09-04
Fix typo in header comment for alloc_copy_buffer()

sjmudd:
Fix trivial minor typo in the header comment.


sjmudd:

sjmudd:
See: [bug#104853](https://bugs.mysql.com/bug.php?id=104853)

ericherman:
Good catch! :-)

PR #368, opened 2021-09-09
Modify 'The table '...' is full message to indicate the configuration setting

sjmudd:
This error message in 8.0 has caused quite a lot of confusion and to the
developer it is not clear why the table is full or what they can do about it.

While the developer can not probably adjust the setting directly providing
the related configuration setting provides a pointer to the limitations
and settings that can be modified if needed.

Ideally the other language translations need to be made.  Is this necessary to accept the PR?
Also the test cases may need adjusting.

If you require to resolve these points please let me know.

The main thing here is that the existing error message is confusing to developers and providing a hint to the configuration setting would help them look up the details and understand the problem better and then resolve it.


sjmudd:
See: [bug#104889](https://bugs.mysql.com/bug.php?id=104889).

sjmudd:

PR #370, opened 2021-10-02
Update function comment for disableMaster() within semisync_master.h

blylei:
Just modify the comment


blylei:
Yes, This step has finish

blylei:

PR #376, opened 2021-10-21
Remove unused variables

erikfroseth:
Remove multiple unused variables as reported by Clang 13


erikfroseth:

PR #379, opened 2021-11-07
fix cmake files for macos 12.0.1

johnrobbet:
The os version is 12.0.1 in macOS Monterey, and the following check of os version will fail when the sub version is 0, which will cause cmake failing on the latest macOS:

NOT CMAKE_MATCH_2

The condition here should be if the sub version is found, so the condtion should be:

NOT DEFINED CMAKE_MATCH_2

cmake generation succeeded after the change was applied.


chrstphrchvz:
I reported the same issue a couple of days ago. The most recent comment is that the same exact fix proposed here has already been made upstream (internally?): https://bugs.mysql.com/bug.php?id=105464#c517341

johnrobbet:

PR #381, opened 2021-11-22
BUG#105290 ON DELETE CASCADE with generated column crashes in innobas…

rahulmalik87:
…e_get_computed_value

 Problem:
    Virtual column template of child table is not populated when cascade
    delete happened on parent table.  Failing to initliaze it causes a crash
    when we are trying remove the secondary index rows defined on virtual column .

    Analysis:
    A cascade delete from the parent table triggers delete on the child table.
    Before a clustered index record is deleted in the child table, a copy of row is
    built to remove secondary index records. This copy of the row requires virtual
    columns to be materialized. Hence, if child table has any virtual columns,
    we have to initialize virtual column template

    Fix:
    As part of fix e5e8c9d1b2e0 MySQL handle case when child virtual column
    depends on forign key. To extend the fix we are now initliazing virtual column template
    if child table has any virtual column.

### DIFF
--- a/mysql-test/suite/innodb/t/foreign_key.test
+++ b/mysql-test/suite/innodb/t/foreign_key.test
@@ -253,3 +253,21 @@ SELECT `date_sent` FROM `email_stats` WHERE `generated_sent_email` = '2020-';
 #clean up.
 DROP TABLE `email_stats`;
 DROP TABLE `emails`;
+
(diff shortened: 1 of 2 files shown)


rahulmalik87:
> Hi, thank you for submitting this pull request. In order to consider your code we need you to sign the Oracle Contribution Agreement (OCA). Please review the details and follow the instructions at https://oca.opensource.oracle.com/
> Thanks

Thanks for revert. OCA is read signed by My Employer Percona.

rahulmalik87:
I confirm the code being submitted is offered under the terms of the OCA, and that I am authorized to contribute it. My public email is rahul.malik@percona.com.

PR #383, opened 2021-12-07
Replace systemd PermissionsStartOnly with executable prefix

mvorisek:
`PermissionsStartOnly` option is deprecated

executable prefix docs: https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=


mvorisek:

mvorisek:
I do not have Oracle account nor want to create any. If merging it is problematic even if I declare I am the author of it, please take it as inspiration only and feel free to commit it under your own name.

mysql-admin:
@mvorisek 
Thanks for your suggestion
==Omer

mvorisek:
there was! :)

mvorisek:
this PR was closed by a bot, can it be merged by someone or is the OCA needed to be signed even for such sudgestion legally?

terjeros:
Hi!

I have transfer PR to bug report  and will have a look:

 https://bugs.mysql.com/bug.php?id=106468

Thanks for reporting!

terjeros:
This breaks on el7 which have:

$ rpm -q systemd
systemd-219-78.0.7.el7.x86_64

While current setup works for all supported platforms. 

Need logic to detect target platform  then, not sure if that make sense at this point?

mvorisek:
I need it at least for Debian, but I am quite quite sure all latest OS distros support the new directives.

PR #385, opened 2021-12-31
libmysql: fix musl compatibility

stintel:
The musl C library does not support res_ninit, res_nsearch and
res_nclose. Check if res_ninit is defined, and use res_search if not.


stintel:
OCA approved and processed.

stintel:

PR #386, opened 2021-12-31
rpcgen.cmake: always build xdr_sizeof.c on macOS

chrstphrchvz:
The availability of `xdr_sizeof()` is dependent on the macOS version, not the compiler version; `xdr_sizeof()` was provided by macOS 10.13 and earlier through the libsystem_info.dylib system library. Rather than adjusting the check to reflect this, xdr_sizeof.c can be built unconditionally on macOS, since MySQL has dropped support for macOS 10.13 and earlier as of 2020-01-13: https://www.mysql.com/support/eol-notice.html.


chrstphrchvz:

PR #391, opened 2022-03-08
BUG#106616 MS 8.0 upgrade (from MS 5.7) crashes with Assertion failur…

rahulmalik87:
…e: dict0dict.cc:1236:table2 == nullptr

Problem:
crash while upgrade due to duplicate table space object.

Analysis:
During upgrade to 8.0, table_id are increased by DICT_MAX_DD_TABLES (1024). The
DICT_MAX_DD_TABLES are reserve for DD tables.
Subjected 5.7 schema has table SYS_FIELDS with table_id 4 and SYS_VIRTUAL
with table_id 1028. Once SYS_FIELDS is upgrade to 4 + 1024 = 1028.
PS asserts while trying to upgrade SYS_ZIP_DICT since server already has
table open with 1028

Fix:
SYS_* tables are special. All SYS_* tables are loaded and then their table
ids are changed. Some SYS tables like SYS_VIRTUAL can have ids > 1024
(say 1028). When changing table_ids of SYS_FIELDS, from 4 to 1028, it will
conflict with the table_id with SYS_VIRTUAL which haven't
been shifted by 1024 yet and currently loaded with 1028. Hence, we change ids
of these SYS tables in reverse order.


rahulmalik87:
I confirm the code being submitted is offered under the terms of the OCA, and that I am authorized to contribute it."
Thanks you

PR #394, opened 2022-03-24
Adding CYCLE timer for s390x

namrata-ibm:
Tests fail with error `The CYCLE timer is not available` on s390x. This PR will fix it.


namrata-ibm:

PR #397, opened 2022-04-21
Fix incorrect value of CACHELINESIZE on s390x

namrata-ibm:
The `sysconf(_SC_LEVEL1_DCACHE_LINESIZE)` call which is used to get the CPU cache line size returns 0 on s390x RHEL 7.x, which is an incorrect value. This causes `rpl_commit_order_queue` and `integrals_lockfree_queue` to fail. This PR is for fixing the same.


namrata-ibm:

PR #398, opened 2022-05-07
Fix path argument supplied to tar in boost.cmake

GBuella:
Unpacking a downloaded Boost package was not possible,
due to supplying erroneous argument to the tar command.

Before:
```
cd directory; tar xfj directory/boost.tar.bz2
```

After:
```
cd directory; tar xfj boost.tar.bz2
```


GBuella:

PR #400, opened 2022-05-16
Update mysql.cc

zzjshaok:
fix: memory leak when mysql client reconnect


zzjshaok:
Hi，ASAN shows  ./bin/mysql has memory leak  when it reconnects

mysql> use test;
No connection. Trying to reconnect...
Connection id:    8
Current database: *** NONE ***

Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> use test;
Database changed
mysql> exit;

PR #402, opened 2022-05-18
fix global-buffer-overflow of fix_paths

yezhiweicmss:
When passing the argument "--basedir=", mysqld with ASAN crashed for global-buffer-overflow in fix_paths.


yezhiweicmss:

PR #406, opened 2022-06-06
Each slow log prints out the database name

diyuhuodong:
MySQL's slow log does not support rotate. So we go to rotate the slow.log by ourselves, but when we analyze the new slow.log separately, we cannot get the db of the sql because there is no use db.
E.g:
slow.log records:
use db;
select * from t1;

rotate slow.log

new slow.log records:
select * from t2;

We can't get the db of that sql from new slow.log.


diyuhuodong:

PR #411, opened 2022-06-16
Bug#107462 Failed upgrade attempt to 8.0.29 corrupts the data dictionary

rahulmalik87:
https://bugs.mysql.com/bug.php?id=107462

Problem:
Failed upgrade attempt to 8.0.29 corrupts the data dictionary
Analysis:
8.0.29 has changed the redo log format and shuffle the redo type code,
so lower version can't parse redo log file in case of failed upgrade
Fix:
Recreate the empty redo log file during failed upgrade
so lower version does not need to parse redo file


rahulmalik87:

PR #412, opened 2022-06-16
BUG-107207 Assertion Failure buf0buf.cc:4318:page_size.equals_to(space…

rahulmalik87:
…_page_size)

https://bugs.mysql.com/bug.php?id=107207

Problem:
Crash in debug build when discard table followed by drop table.

Analysis:
As part of discard tablespace, server sets index root page to null but
does not remove AHI(Adaptive Hash Index) entries from buffer pool.It can
causes crash if there a drop after the discard tablespace. As drop
tablespace explicity remove AHI entries but these indexes root pages are
already set to null and it asserts different places in code

Fix:
drop AHI as part of discard tablespace


rahulmalik87:

PR #415, opened 2022-07-05
Comment explain why not protect shared-variable

hawkxiang:
'srv_thread_sleep_delay' maybe change by multi threads. 
Add Comments to explain why this variable need't protect.


hawkxiang:
I submitted the OCA document.

PR #420, opened 2022-09-06
Implement logic to be able to get a consistent GTID

dbussink:
This adds logic so that with a `START TRANSACTION WITH CONSISTENT SNAPSHOT`, we can also get a correct GTID that is appropriate at the time the transaction is started.

It locks around the logic when a consistent snapshot is created in the underlying storage engine and grabs the current GTID under lock at that point.

That means that no other threads can update the GTID at the exact same point in time avoiding the race that currently exists if you execute `SELECT @@global.gtid_executed` which doesn't enforce such locking.

The locking overhead is should be relatively small. When using a consistent read snapshots, systems (like Vitess) often want to get a consistent GTID. Today that's only possible by then also locking all tables which is much more expensive.

The implementation here opts to use a new specific `START_GTID` option. There are other options though here. I can also see the argument for if we'd want to return when `OWN_GTID` or `ALL_GTID` is enabled to also return this on starting transactions this way. There could be a compatibility concern there since a client might not expect a GTID in the `OK` packet on a transaction start. It should still be able to parse a generic `OK` packet though, so maybe this is fine.

I'm all open for changing that up if that is preferred instead (and we can update our PlanetScale fork also accordingly since we'd like to stick as close to mainline as we can :smile:). 

cc @lefred since you asked in https://github.com/planetscale/mysql-server/commit/9fa67ec3ebdeda923a67963260a9e744aa635a3a#commitcomment-82813813

### DIFF
--- a/sql/handler.cc
+++ b/sql/handler.cc
@@ -2326,7 +2326,17 @@ static bool snapshot_handlerton(THD *thd, plugin_ref plugin, void *arg) {
 int ha_start_consistent_snapshot(THD *thd) {
   bool warn = true;
 
+  /*
+    Grabbing the write lock here since we end up calling
+    store_start_gtid => notify_start_gtid_determined => ensure_sidno
+    which wants the write lock to potentially expand storage for sidnos.
+  */
+  global_sid_lock->wrlock();
+  thd->rpl_thd_ctx.session_gtids_ctx().notify_start_gtid_determined(thd);
+
   plugin_foreach(thd, snapshot_handlerton, MYSQL_STORAGE_ENGINE_PLUGIN, &warn);
+
+  global_sid_lock->unlock();
 
   /*
     Same idea as when one wants to CREATE TABLE in one engine which does not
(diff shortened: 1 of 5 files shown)


shlomi-noach:
Cross referencing https://github.com/vitessio/vitess/pull/11061#issuecomment-1222117793 ; the new behavior has been tested in Vitess against a patched MySQL, in a stress test for VReplication; the test is designed to fail on minuscule mistakes in GTID evaluation, and was able to invalidate multiple other failed attempts at reaching consistent transactional GTIDs. The patch offered in this PR passes that stress test flawlessly and consistently.

dbussink:

shlomi-noach:
Looks a bit unexpected that the bot closed this PR?

mysql-admin:
Hi @shlomi-noach,
This has always been the case with contributions. 
Once contributions are accepted we transfer them to bugs in bugs.mysql.com  (in this case bug#108405) and work continues on them as bug fixes/enhancements. Updates are reflected in the bug record (not the PR) so there is no point in keeping the PR open

Thanks
==Omer

PR #421, opened 2022-09-06
fix some error handlers in mysqlpump

yezhiweicmss:
1. ferror(m_file) when fclose(m_file) error;
2. compression_writer_as_wrapper NULL dereferenced in"Unknown compression method";
3. dump_task NULL checked and dereferenced;
4. dynamic_cast<Abstract_data_object*>(new_priv_task)) maybe NULL but not checked.


yezhiweicmss:

PR #422, opened 2022-09-06
Fix auth_sec.keyring_file_data_qa test in docker

hatvik:
Running this test in docker gives an additional error in the log:
`2022-08-23T11:31:00.065464Z 17 [ERROR] [MY-011370] [Server] Plugin keyring_file reported: 'Cannot seek in file '/' (OS errno 22 - Invalid argument)'
`

To reproduce:
`./mtr auth_sec.keyring_file_data_qa`
or
```
docker run -it -d --name mybug -e MYSQL_ROOT_PASSWORD=my-secret-pw mysql/mysql-server:latest
docker exec -it mybug mysql -p'my-secret-pw' -e'INSTALL PLUGIN keyring_file SONAME "keyring_file.so"';
docker exec -it mybug mysql -p"my-secret-pw" -e"SET @@global.keyring_file_data='/'";
docker logs mybug
```

Fix is to suppress the additional log message.


hatvik:

PR #427, opened 2022-10-25
Bug #108870: Eliminated dead code discovered with clang-15 compiler

percona-ysorokin:
https://bugs.mysql.com/bug.php?id=108870

Fixed a number of "unqualified call to 'std::move'" warnings ('-Wunqualified-std-cast-call').

Along with that, removed 'using std::move' from the source files.

Fixed "a function declaration without a prototype is deprecated in all versions of C" warnings ('-Wstrict-prototypes') in 'libcbor'.

Fixed problem with detecting 'std::is_trivially_copyable' presence in 'robin-hood-hashing/robin_hood.h' when building with a 'clang' family of compilers.

Removed a lot of dead code detected with new 'clang-15' optimizer ('-Wno-unused-but-set-variable').

Binaries compiled with 'clang-15' under Address Sanitizer ('-DWITH_ASAN=ON -DWITH_ASAN_SCOPE=ON') can no longer be run with 'thread_stack' set to '1M'. For this combination of compiler / configuration options we now set the default value of this system variable to '4M'

Fixed a number of 'out-of-line definition of constexpr static data member is redundant in C++17 and is deprecated' warnings by removing these redundant definitions.

Because of that, removed 'base64.cc', 'content_type.cc', 'internet.cpp', and 'socket.cpp' from  the 'router' as they became empty source files.

'harness_net_ts' target converted to header-only library as it no longer has '.cc' files.

Reworked 'json_schema_embedder': it no longer needs to generate an empty '.cc' file (without 'constexpr' definition ) - an '.h' is now everything that is needed. Because of this, the utility now expects just 3 command line arguments '<in_file> <hdr_file> <symbol_name>'. 'CMakeLists.txt' files invoking 'json_schema_embedder' updated correspondingly.

Suppressed '-Wno-unused-but-set-variable' warning in Bison-generated files ('sql_yacc.cc', 'sql_hints.yy.cc', and 'fts/fts0pars.cc').

Due to partial support of DWARF-5 debug info format in Valgrind 3.18 (the latest out-of-the-box version available on supported platforms), in order to run MySQL Server 8.0.31 under Valgrind it has to be configured with the following CMake options
-DCMAKE_C_FLAGS_DEBUG=-gdwarf-4
-DCMAKE_CXX_FLAGS_DEBUG=-gdwarf-4
-DCMAKE_C_FLAGS_RELWITHDEBINFO="-O2 -gdwarf-4 -DNDEBUG" -DCMAKE_CXX_FLAGS_RELWITHDEBINFO="-O2 -gdwarf-4 -DNDEBUG" Consider, including this logic into 'compiler_options.cmake'.


percona-ysorokin:
The OCA-signed patch has already been submitted to
https://bugs.mysql.com/bug.php?id=108870

percona-ysorokin:

PR #430, opened 2022-11-13
Fix incorrect suggested commands to start slave in error logs.

dmccombs:
I noticed when running into a duplicate key error during slave replication today that the command listed in the error log to start the slave thread is backwards, saying `SLAVE START`, when the correct command is actually `START SLAVE`.


dmccombs:
I've completed the contribution agreement form, its status is currently Under Review.

dmccombs:

PR #433, opened 2022-12-14
BUG#32959894

mronstro:
ClusterJ uses a scratch buffer for primary key hash calculations which is limited to 10000 bytes. This is too small in some cases.

The solution is to malloc the buffer in computeHash if the buffer isn't big enough. This also solves some application problems.

This patch contains a test for this bug that ensures that the primary key requires a buffer larger than 10000 bytes.


mronstro:

PR #435, opened 2022-12-15
cmake: fail fatally when sql/sql_yacc.h is missing and bison is not present

jelmer:
This ensures that the configure step fails, rather than halfway through the build.


jelmer:

PR #436, opened 2023-01-03
Fix some startup problems in mysqld_pre_systemd when using multiple MySQL Server with systemd

JackFlame:
There are some logic problems in original script . The script supposes user doesn't define `datadir` and `log-error` , and gives them default values , but when starting MySQL with **systemd** , the mysqld.service and **mysqld@.service** doesn't use the default values , which cause server startup errors . I modified this script by adding these options with default values into corresponding option group in **/etc/my.cnf** . So **mysqld** program can start successfully using the default values of these options which define by the **mysqld_pre_systemd** script . And sometimes MySQL Server startup failed but doesn't log any errors in `log-error` instead of printing in **stdout** . So I add `--log-error` command option to those command to avoid that .

### DIFF
--- a/scripts/systemd/mysqld_pre_systemd.in
+++ b/scripts/systemd/mysqld_pre_systemd.in
@@ -23,17 +23,18 @@
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA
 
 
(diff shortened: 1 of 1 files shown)


JackFlame:
I already sign OCA . So what should I do next ?

PR #437, opened 2023-01-11
BUG-109433 mysqld crash when IF more than 64

zzt93:
To fix [BUG-109433](https://bugs.mysql.com/bug.php?id=109433)

Problem:
Crash when network interface is more than 64 (when deploy with k8s, who create many virtual network interface)

Analysis:
`ifrp` is overflowed when network interface is more than IFRP_INIT_ARR_SIZE(64). It is not reallocated because wrong condition `i==ifrpsize`.

Fix:
fix wrong reallocate condition.


zzt93:
@mysql-oca-bot I was added under Bytedance OCA. My MySQL bug system user email: zengzetang@bytedance.com

zzt93:

PR #443, opened 2023-02-07
client/mysql: enable comments by default

dveeden:
Make `--comments` the default for MySQL Client.

Stripping comments was very useful to make the MySQL Query Cache get a higher hitrate. However with the query cache having been deprecated in 5.7.20 and removed in 8.0 this isn't needed anymore.

Having the comments sent to the server make them appear in the processlist which can make them more easy to identify. Making them available to routers and proxies allows those to make use of them for query routing and diagnostic purposes as well.


dveeden:

PR #448, opened 2023-03-22
return error immediately after flush_thread_caches failed.

lishuai87:
Problem:
--------
MYSQL_BIN_LOG::ordered_commit() flush all thread caches to binlog file by calling MYSQL_BIN_LOG::process_flush_stage_queue(). If one thread flush_thread_caches fail, the binlog file may have corrupted data. If another thread continue flush_thread_caches, it may success, then MYSQL_BIN_LOG::process_flush_stage_queue() return success. Then mysql commit success, with corrupted binlog data.

Solution:
---------
Return error immediately after flush_thread_caches failed.


lishuai87:
https://bugs.mysql.com/bug.php?id=110465

PR #449, opened 2023-03-28
Fix crashing mysql due to inadequate cleanup of cached queries that use TempTable

lottaquestions:
Description
============
The TempTable storage engine can cause the MySQL engine to crash
if the same query that created a temporary table is run more than
once. The reason for this crash is that the TempTable handler is
cleaned up after execution of the query, but the "keys in use" flag
for the underlying temporary table remains "on" in the cached results.
When the query is rerun, it uses the cached results from the
previous run, and since the "keys in use" flag is set, logic in the
execute stage of the query attempts to access a pointer to what it
expects is an opened temporary table. But since the TempTable
handler had been cleaned up after the first run, a new TempTable
handler is instantiated for the new run, which does not yet have
a pointer to an opened temporary table. When the query rerun
attempts to access the opened temporary table via a null pointer
the system crashes. The crashes only happen on the second
rerun of the query. Subsequent reruns after the second one do
not crash the engine.

This change fixes this issue by ensuring that the "keys in use" flag
is reset (0) during the TempTable handler clean up, after the first
run of a query. This is done because, if the underlying
temporary table has been cleaned up, then by definition there should
not be any "keys in use" on it.

Fixes bug https://bugs.mysql.com/bug.php?id=110152 reported to
Oracle MySQL.

Testing
========
An MTR test is provided with the PR that forces the creation of 
temporary tables and hence causes a crash without the change, but
works correctly with the change.

### DIFF
--- a/mysql-test/r/temptable_repeat_cached_query.result
+++ b/mysql-test/r/temptable_repeat_cached_query.result
@@ -1,0 +1,27 @@
+DROP DATABASE IF EXISTS db1;
+CREATE DATABASE `db1`;
+USE `db1`;
+CREATE TABLE db1.`tb1` (
+`col1` varchar(8) NOT NULL 
+);
+INSERT INTO db1.`tb1`(col1) VALUES ('20211114');
+CREATE PROCEDURE db1.`stored_proc1`()
+BEGIN
+WITH t_dataset_info as (
+SELECT col1 as yyyymmdd FROM tb1 WHERE col1 = (SELECT max(col1) FROM tb1 )
+)
+SELECT a.yyyymmdd as yyyymmdd
+FROM (
+SELECT a.yyyymmdd
+FROM t_dataset_info a
+WHERE a.yyyymmdd = (SELECT max(a.yyyymmdd) FROM t_dataset_info a WHERE a.yyyymmdd <= '20220131')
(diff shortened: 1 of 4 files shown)


mysql-admin:
==Omer

lottaquestions:

lottaquestions:
> Hi, thank you for your contribution. Your code has been assigned to an internal queue. Please follow bug http://bugs.mysql.com/bug.php?id=111465 for updates. Thanks

I am unable to view the contents in the above mentioned bug report even when logged into my Oracle account. Please give me access.

mysql-admin:
Hi @lottaquestions 
See reply in direct email 

Thanks
==Omer

lottaquestions:
Fixed in MySQL 8.0.35. See https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-35.html#mysqld-8-0-35-bug :+1: 

> 
> Some cached queries were not always handled correctly.
> 
> Our thanks to Nicholas Othieno for the contribution. (Bug #35105404, Bug #35511208)

Commit that fixed the issue is https://github.com/mysql/mysql-server/commit/a1d235e83859ed39610b2d39eb1a7642964de72c

PR #450, opened 2023-04-03
mysql 5.7 audit plugin add examined_row_count/affected_row_count/return_row_count parameter

hongdongjian:
audit plugin add examined_row_count/affected_row_count/return_row_count parameter

PR #451, opened 2023-04-03
mysql 8.0 audit plugin add examined_row_count/affected_row_count/return_row_count parameter

hongdongjian:
for mysql 8.0, audit plugin add examined_row_count/affected_row_count/return_row_count parameter

### DIFF
--- a/include/mysql/plugin_audit.h
+++ b/include/mysql/plugin_audit.h
@@ -141,6 +141,9 @@ struct mysql_event_general {
   MYSQL_LEX_CSTRING general_sql_command;
   MYSQL_LEX_CSTRING general_external_user;
   MYSQL_LEX_CSTRING general_ip;
+  unsigned long long general_examined_row_count;
+  unsigned long long general_affected_row_count;
+  unsigned long long general_return_row_count;
 };
 
 #define MYSQL_AUDIT_CONNECTION_ALL                                      \
--- a/include/mysql/plugin_audit.h
+++ b/include/mysql/plugin_audit.h
@@ -141,6 +141,7 @@ struct mysql_event_general {
   MYSQL_LEX_CSTRING general_sql_command;
   MYSQL_LEX_CSTRING general_external_user;
   MYSQL_LEX_CSTRING general_ip;
+  MYSQL_LEX_CSTRING general_database;
   unsigned long long general_examined_row_count;
   unsigned long long general_affected_row_count;
   unsigned long long general_return_row_count;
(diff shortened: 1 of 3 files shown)


hongdongjian:

gkodinov:
Thank you for your attempt at contributing this enhancement. While it's a valid need, unfortunately this is not the way to implement it. Your change breaks ABI (application binary interface) around the audit plugin (as exemplified by the need to change the .pp file). Plus, it adds the new values for all audit plugins, regardless whether they need these or not.

I would suggest you consider turning the "push" model you have come up now into a "pull" model and extend e.g. https://dev.mysql.com/doc/dev/mysql-server/latest/structs__mysql__mysql__thd__attributes.html with the data you need.

hongdongjian:
> Thank you for your attempt at contributing this enhancement. While it's a valid need, unfortunately this is not the way to implement it. Your change breaks ABI (application binary interface) around the audit plugin (as exemplified by the need to change the .pp file). Plus, it adds the new values for all audit plugins, regardless whether they need these or not.
> 
> I would suggest you consider turning the "push" model you have come up now into a "pull" model and extend e.g. https://dev.mysql.com/doc/dev/mysql-server/latest/structs__mysql__mysql__thd__attributes.html with the data you need.

Thank you for your reply and guidance. You are right. This is my new submission [https://github.com/mysql/mysql-server/pull/453]. Would you mind taking a look when you have time?

hongdongjian:
> Thank you for your attempt at contributing this enhancement. While it's a valid need, unfortunately this is not the way to implement it. Your change breaks ABI (application binary interface) around the audit plugin (as exemplified by the need to change the .pp file). Plus, it adds the new values for all audit plugins, regardless whether they need these or not.
> 
> I would suggest you consider turning the "push" model you have come up now into a "pull" model and extend e.g. https://dev.mysql.com/doc/dev/mysql-server/latest/structs__mysql__mysql__thd__attributes.html with the data you need.

Hello, I would like to know if there is a similar way to obtain these variables from THD in version 5.7. This is regarding my previous submission [https://github.com/mysql/mysql-server/pull/450].

gkodinov:
done. please try to stick to the same pull request if at all possible.

gkodinov:
Sorry, but nothing like this in 5.7. 5.7 is to be going out of active development pretty soon. Let's focus on 8.0 :)

hongdongjian:
> done. please try to stick to the same pull request if at all possible.

Thank you for your reply. I will take some time later to revise it.

hongdongjian:
> Sorry, but nothing like this in 5.7. 5.7 is to be going out of active development pretty soon. Let's focus on 8.0 :)

understand

PR #453, opened 2023-04-07
THD Attributes register db/examined_row_count/row_count_func/sent_row_count

hongdongjian:
THD Attributes register m_db/m_examined_row_count/m_row_count_func/m_sent_row_count

### DIFF
--- a/sql/server_component/mysql_thd_attributes_imp.cc
+++ b/sql/server_component/mysql_thd_attributes_imp.cc
@@ -50,6 +50,17 @@ DEFINE_BOOL_METHOD(mysql_thd_attributes_imp::get,
         *((bool *)inout_pvalue) = t->is_server_upgrade_thread();
       } else if (!strcmp(name, "is_init_file_thread")) {
         *((bool *)inout_pvalue) = t->is_init_file_system_thread();
+      } else if (!strcmp(name, "db")) {
+        if (t->db().str == nullptr) return true;
+        String *res = new String[1];
+        res->copy(t->db().str, t->db().length, &my_charset_utf8mb3_bin);
+        *((my_h_string *)inout_pvalue) = (my_h_string)res;
+      } else if (!strcmp(name, "examined_row_count")) {
+        *((ulonglong *)inout_pvalue) = t->get_examined_row_count();
+      } else if (!strcmp(name, "row_count_func")) {
+        *((longlong *)inout_pvalue) = t->get_row_count_func();
+      } else if (!strcmp(name, "sent_row_count")) {
+        *((ulonglong *)inout_pvalue) = t->get_sent_row_count();
       } else
         return true; /* invalid option */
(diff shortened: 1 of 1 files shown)


sql/server_component/mysql_thd_attributes_imp.cc
```diff
@@ -50,6 +50,14 @@ DEFINE_BOOL_METHOD(mysql_thd_attributes_imp::get,
         *((bool *)inout_pvalue) = t->is_server_upgrade_thread();
       } else if (!strcmp(name, "is_init_file_thread")) {
         *((bool *)inout_pvalue) = t->is_init_file_system_thread();
+      } else if (!strcmp(name, "m_db")) {
+        *((LEX_CSTRING *)inout_pvalue) = t->db();
```
gkodinov: it doesn't work this way: LEX_CSTRING doesn't for example carry charset info. You're better off using mysql_string strings (this is the official charset compatible way to return data) or you need to convert to a known charset (utf-8) and copy it in a buffer.

sql/server_component/mysql_thd_attributes_imp.cc
```diff
@@ -50,6 +50,14 @@ DEFINE_BOOL_METHOD(mysql_thd_attributes_imp::get,
...
+      } else if (!strcmp(name, "m_examined_row_count")) {
+        *((ha_rows *)inout_pvalue) = t->get_examined_row_count();
```
gkodinov: ha_rows is an internal type. Please use one of the C types that's closest. e.g. an unsigned long long.

sql/server_component/mysql_thd_attributes_imp.cc
```diff
@@ -50,6 +50,14 @@ DEFINE_BOOL_METHOD(mysql_thd_attributes_imp::get,
         *((bool *)inout_pvalue) = t->is_server_upgrade_thread();
       } else if (!strcmp(name, "is_init_file_thread")) {
         *((bool *)inout_pvalue) = t->is_init_file_system_thread();
+      } else if (!strcmp(name, "m_db")) {
+        *((LEX_CSTRING *)inout_pvalue) = t->db();
+      } else if (!strcmp(name, "m_examined_row_count")) {
```
gkodinov: don't to the "m_" prefix in the names please.

gkodinov (changes requested):
I'd also add (empty) switch cases in set to signify that these are taken but aren't settable. 
And I'd consider some mtr or unit test.

hongdongjian:

hongdongjian:
> I'd also add (empty) switch cases in set to signify that these are taken but aren't settable. And I'd consider some mtr or unit test.

Thank you for your guidance. I have made the corresponding changes according to your advice. When you have time, could you please help me take a look and I look forward to your reply.

PR #455, opened 2023-04-09
sql/memory: Fix build on musl

thesamesam:
_SC_LEVEL1_DCACHE_LINESIZE is not always available on Linux, e.g. with musl libc.

It's provided by glibc instead and is a glibc-internal.


thesamesam:

PR #456, opened 2023-04-12
sql/binlog/group_commit: include header for std::uint64_t

parona-source:
GCC 13 changed default header dependencies for C++ necessitating including cstdint explicitly.


blaudden:
This has been fixed in an upcoming release of MySQL Server. Thanks for noticing!

parona-source:
Good to hear :+1:

PR #463, opened 2023-05-05
Protocol docs: zstd level is missing an if statement

dveeden:
The `zstd_compression_level` is only sent when `CLIENT_ZSTD_COMPRESSION_ALGORITHM` is set.

See also:
- `parse_client_handshake_packet()`


dveeden:

PR #465, opened 2023-05-08
CONTRIB_371: Efficient zeroing of UNDO, REDO and tablespace files

mronstro:
When initialising REDO log files at startup we want to ensure that the entire file size is preallocated and that the file is properly initialised.

The same is true for UNDO log files in LGMAN and tablespace files.

To speed up this we use special XFS methods and POSIX methods that apply to the file systems we recommend users to use. This means that we can preallocate file space without actually writing it and we can even ensure that reads of this area returns pages filled with zeroes.

We can also use fallocate which works for XFS and ext4 at least in Linux.

We still need to write the initial parts of the file and for the REDO log we have to write each start page of each MByte.

Encrypted and compressed files have a file header that should not be zeroed. More debug info.

### DIFF
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -1,4 +1,5 @@
 /* Copyright (c) 2009, 2023, Oracle and/or its affiliates.
+   Copyright (c) 2023, 2023, Hopsworks and/or its affiliates.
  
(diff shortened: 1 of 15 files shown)


mronstro:

Rgrds Mikael Ronström

PR #466, opened 2023-05-10
fix: wheather field data needs convert when field ddl change from utf…

mswdwk:
…8 to utf8mb4

### DIFF
--- a/sql/field.cc
+++ b/sql/field.cc
@@ -8000,7 +8000,7 @@ uint Field_varstring::is_equal(Create_field *new_field)
 {
   if ( (new_field->sql_type == real_type() &&
         new_field->charset == field_charset) || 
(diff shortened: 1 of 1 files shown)


sql/field.cc
```diff
@@ -7998,8 +7998,11 @@ Field *Field_varstring::new_key_field(MEM_ROOT *root,
 
 uint Field_varstring::is_equal(Create_field *new_field)
 {
-  if (new_field->sql_type == real_type() &&
-      new_field->charset == field_charset)
+  if ( (new_field->sql_type == real_type() &&
+        new_field->charset == field_charset) || 
+       (new_field->chaset == &my_charset_utf8mb4_bin &&
```
dlenski: This contains a typo and won't compile, so it clearly hasn't been tested.

PR #468, opened 2023-05-23
Set variables to PARENT_SCOPE in CMake functions

keeper:
1f2b9d62 replaces macros with functions, but in CMake functions have different scope and the variables set in the functions are not visible to the caller. The change breaks MySQL building with `WITH_ZLIB="system"` for the reason above (ZLIB_FOUND is not visible when returning from `FIND_SYSTEM_ZLIB()`).

Setting the necessary variables to PARENT_SCOPE (equal to returning variables) solves this issue.


keeper:

dlenski:
After going through the Magical Internal Oracle/MySQL Blender™, this was eventually merged in MySQL 8.0.35 as https://github.com/mysql/mysql-server/commit/486fb151d050200cae8bb3a3a65b273306ede386

PR #471, opened 2023-05-26
Replace MySQL Internals Manual links

dveeden:
The MySQL Internals Manual links now also use the same generic redirect instead of redirecting to the specific information that it was pointing to before. This tries to update the majority of the links that were still in the source code.


dveeden:

PR #485, opened 2023-07-21
Architecture-agnostic fix for crashes caused by cache coherency_line_size of 0 in sysfs

dlenski:
Previous fixes for this issue have been piecemeal and CPU architecture-specific:

- For `s390x`: BUG#34095278, https://bugs.mysql.com/bug.php?id=107081 → f2d64bfa575700234d5f36418d5ef143b54e2a16 for s390x
- For `armhf`: https://bugs.mysql.com/bug.php?id=110752
- All other arm variants are also affected (confirmed by us for `arm64`)

It would be far better to address this in a proactive, architecture-agnostic way, to prevent future cases on unexpected CPU architecture or ABI variants.

We can do this by making the `_cache_line_size()` function check for invalid values in sysfs (`/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size`), and fall back to 64 as a default return value.

All new code of the whole pull request, including one or several files that are either new files or modified ones, are contributed under the BSD-new license. I am contributing on behalf of my employer Amazon Web Services

### DIFF
--- a/sql/memory/aligned_atomic.h
+++ b/sql/memory/aligned_atomic.h
@@ -80,19 +80,21 @@ static inline size_t _cache_line_size() {
 #elif defined(__linux__)
 static inline size_t _cache_line_size() {
   long size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE);
-  if (size == -1) return 64;
-#if defined(__s390x__)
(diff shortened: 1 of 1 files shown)


dlenski:

PR #488, opened 2023-08-03
Bug #109668: XCom connection stalled forever in read() syscall over network

venkatesh-prasad-v:
https://bugs.mysql.com/bug.php?id=109668

Description
-----------
GR suffered from problems caused by the security probes and network scanner processes connecting to the group replication communication port. This usually is not a problem, but poses a serious threat when another member tries to join the cluster by initialting a connection to the member which is affected by external processes using the port dedicated for group communication for longer durations.

On such activites by external processes, the SSL enabled server stalled forever on the SSL_accept() call waiting for handshake data. Below is the stacktrace:

    Thread 55 (Thread 0x7f7bb77ff700 (LWP 2198598)):
    #0 in read ()
    #1 in sock_read ()
    #2 in BIO_read ()
    #3 in ssl23_read_bytes ()
    #4 in ssl23_get_client_hello ()
    #5 in ssl23_accept ()
    #6 in xcom_tcp_server_startup(Xcom_network_provider*) ()

When the server stalled in the above path forever, it prohibited other members to join the cluster resulting in the following messages on the joiner server's logs.

    [ERROR] [MY-011640] [Repl] Plugin group_replication reported: 'Timeout on wait for view after joining group'
    [ERROR] [MY-011735] [Repl] Plugin group_replication reported: '[GCS] The member is already leaving or joining a group.'

Solution
--------
This patch adds two new variables

1. group_replication_xcom_ssl_socket_timeout

   It is a file-descriptor level timeout in seconds for both accept() and SSL_accept() calls when group replication is listening on the xcom port. When set to a valid value, say for example 5 seconds, both accept() and SSL_accept() return after 5 seconds. The default value has been set to 0 (waits infinitely) for backward compatibility. This variable is effective only when GR is configred with SSL.

2. group_replication_xcom_ssl_accept_retries

   It defines the number of retries to be performed before closing the socket. For each retry the server thread calls SSL_accept()  with timeout defined by the group_replication_xcom_ssl_socket_timeout for the SSL handshake process once the connection has been accepted by the first accept() call. The default value has been set to 10. This variable is effective only when GR is configred with SSL.

Note:
- Both of the above variables are dynamically configurable, but will become effective only on START GROUP_REPLICATION.
- This patch is only for the Linux systems.

### DIFF
--- a/mysql-test/include/start_proc_in_background.inc
+++ b/mysql-test/include/start_proc_in_background.inc
@@ -1,0 +1,71 @@
+# ==== Purpose ====
+#
+# Start process in background. Optionally store its PID in the file
+#
+# This functionality cannot be achieved by mysqltest built-in --exec_in_background
+# because we need to start it like:
+# cmd param1 param2 > output.log & echo $! > cmd.pid
+#
+# The way to pass $! through Perl to mysqltest binary is to escape $! like \$!
+# This will prevent attemption to resolve $! variable on Perl layer,
+# but will pass down \$! to mysqltest.
+# mysqltest handles \$! literally (recognizes that $ has been escaped
+# and adds escape character to the constructed command, so we end up with:
+# cmd param1 param2 > output.log & echo \$! > cmd.pid
+#
+# ==== Usage ====
+#
+# --let $command = process_to_be_executed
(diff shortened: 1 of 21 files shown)


venkatesh-prasad-v:

PR #490, opened 2023-08-17
Bug#110127: semi-sync master wait for ack when semi-sync replica is down & async replaces it & async replaces it

venkatesh-prasad-v:
https://bugs.mysql.com/bug.php?id=110127

Description
-----------
This is the regression introduced in the commit 650d2f7 (Bug#33534218 : replication_sender_observe_commit_only=ON leads to break semisync replication) which made the before_send_hook and after_send_hook to be never called if the event is in the excluded group.

Prior to 8.0.31, dump thread performed below operations
1. Read events from binlog
2. For every event 2.1. Call before send hook 2.2. If the event is skipped 2.2.1. Send hearbeat event if necessary with the latest position 2.2.2. Update the exclude_end_pos to mark the position in binlog till where the replica receiver is synchronized. 2.3. If the event is not skipped 2.3.1. Send heartbeat event if necessary with the last exclude_group_end_pos. 2.3.2. Call send_packet() to send events to replica. 2.4. Call after send hook

This way it ensured that before_send and after_send hooks are executed irrespective of whether the event is skipped or not.

In 8.0.31 upto 8.0.34,
1. Read events from binlog
2. For every event 2.1. If the event is skipped 2.2.1. Send hearbeat event if necessary with the latest position 2.2.2. Update the exclude_end_pos to mark the position in binlog till where the replica receiver is synchronized. 2.2. If the event is not skipped 2.3.1. Send heartbeat event if necessary with the last exclude_group_end_pos. 2.3.2. Call before send hook 2.3.3. Call send_packet() to send events to replica. 2.3.4. Call after send hook

Here, the above upstream commit made the before_send_hook and after_send_hook to be called only when an event is not skipped.

However this change caused the client connection thread to wait for the ACK from semi-sync replica even after it was added back as a semi-sync replica, as this change never reads the acknowledgement from the replica (after send hook is not executed).

Solution
--------
Make the before_send and after_send server hooks to be called even for heartbeat events, preserving the behavior before to 8.0.31, by also preserving the original bugfix made in the commit 650d2f7, i.e, Observe_transmission_guard is created just before calling before_send_hook, so that reserve_header_hook is always called when sending a heartbeat event and magic number is properly encoded.


venkatesh-prasad-v:

PR #491, opened 2023-08-18
fix: function names typo

0o001:


0o001:
I wasn't sure about the 'comit' word , so I reverted it. However, if you find that it's correct, I can commit it again when you check. Thank you.

PR #493, opened 2023-08-24
Bug#111594: add pading to solve false sharing problem

liuchangyan:
Signed-off-by: Teresaliu <changyan.liu@intel.com>
Signed-off-by: Lin Yang <lin.a.yang@intel.com>

## Description

When using sysbench to benchmark mysql's performance,  we found MySQL has some scalability issues as the the number of threads increase in select random ranges/points senarios; more details is there: https://bugs.mysql.com/bug.php?id=111594

After investigation, we located that this problem is a ``false sharing`` problem,  and this kind cacheline contention problem could be reduced by adding code ``pading`` to separate variables in the one cacheline.

## Conclusion
After adding pading, the benchmark results and the performance improvement ratio are as shown in the figure below：
image
image

### DIFF
--- a/storage/innobase/include/buf0buf.h
+++ b/storage/innobase/include/buf0buf.h
@@ -1614,6 +1614,9 @@ class buf_page_t {
    the truncation number. */
   uint32_t m_version{};
 
+  /** Pading to solve false sharing problem */
+  char pading[ut::INNODB_CACHE_LINE_SIZE];
+
   /** Time of first access, or 0 if the block was never accessed in the
   buffer pool. Protected by block mutex */
   std::chrono::steady_clock::time_point access_time;


liuchangyan:

liuchangyan:

mysql-admin:
Hi @liuchangyan 
The contribution code in this PR ([9ba6091](https://github.com/mysql/mysql-server/pull/493/commits/9ba60915c40c99578c72cec8e268079461fdf2b0) )was submitted directly to bug
https://bugs.mysql.com/bug.php?id=111594 (see https://bugs.mysql.com/file.php?id=32551&bug_id=111594&text=1)
so there is n need to submit it again. As such,I am closing the PR.

Thanks for contributing to MySQL . 
==Omer

PR #494, opened 2023-08-28
Fix sloppy home directory substitution in SSL tests

dlenski:
The MTR test files `ssl_ca.test` and `ssl_crl.test` both attempt to verify the absence of a bug relating to Unix home directory substitution, which involves replacing `~` as the leftmost pathname component with the value of the `HOME` environment variable.

However, the tests do so by replacing *all* occurrences of the value of `HOME` with '~' in the absolute path `MYSQL_TEST_DIR`.

This causes problems on systems where the value of `HOME` is a symlink.  For example, if `HOME=/home/myusername` (a symlink to `/local/home/myusername`) and `PWD=/local/home/myusername/MySQL`, the test will end up attempting to use nonsensical broken paths such as `/local~/MySQL/*`.

The solution is to ensure that the substitution occurs *only* at the beginning of the path, and when immediately followed by '/'.

### DIFF
--- a/mysql-test/t/ssl_ca.test
+++ b/mysql-test/t/ssl_ca.test
@@ -16,7 +16,7 @@
 --echo #               PATH SUBSTITUTION
 --echo #
 
---let $mysql_test_dir_path= `SELECT REPLACE('$MYSQL_TEST_DIR', '$HOME', '~')`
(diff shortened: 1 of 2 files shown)


dlenski:
Submitted under the terms of the Amazon MySQL OCA.

dlenski:

PR #497, opened 2023-10-16
Fix usage of BIO_get_mem_data

samuel40791765:
According to the function signature in OpenSSL's documentation, the `BIO_get_mem_data` function should take a `char **` argument instead of a `const char **`. `const char**` and `char **` aren't directly interchangeable, and using `const char**` is considered a misusage.

Reference: https://www.openssl.org/docs/man1.1.1/man3/BIO_get_mem_data.html

I am contributing on behalf of my employer Amazon Web Services.


samuel40791765:

PR #498, opened 2023-10-20
Rondb 475

mronstro:


blaudden:
Problem: The MySQL Server fails when replicating "GRANT NDB_STORED_USER ..." with replication filter turned on. This  occurs since the replication filter causes all non-updating queries to return an error - with the
assumption is that only changes need to be replicated.

We opened  https://bugs.mysql.com/bug.php?id=112775 for the previous contribution and I believe the fix which has been implemented for that bug will provide a generic solution to the problem. 

Fix is to replace the Rpl_filter with an empty Rpl_filter when querying INFORMATION_SCHEMA (and other tables) as otherwise non-updating queries are not allowed in the replica.

It should solve the problem reported by this contribution unless I miss something.

blaudden:
I.e same as https://github.com/mysql/mysql-server/pull/496

svenssonaxel:
> I.e same as #496

@blaudden Correct, and the test case is the exact same. #496 only differs in base (8.1 rather than 8.0) and that the bugfix does not mess with `m_thd->slave_thread`.

PR #502, opened 2023-11-03
Doc: Update compression protocol information

dveeden:
- The "Compressed Packet Header" section says the header is 7 bytes, but only describes the first 4 bytes.
- Clarify what happens if both `CLIENT_COMPRESS` and `CLIENT_ZSTD_COMPRESSION_ALGORITHM` are set.
- Add links to Zstandard website and update text to clarify that there now are multiple compression algorithms.


dveeden:

PR #505, opened 2023-11-15
libmysql: workaround missing res_n* functions on e.g. musl

thesamesam:
musl doesn't have res_n* functions so we are falling back to the not thread safe ones.

Note that another approach was submitted at https://bugs.mysql.com/bug.php?id=106034 (https://github.com/mysql/mysql-server/pull/385) but that is arguably less portable given it takes the new path for everything non-glibc where things may already work fine (e.g. Solaris).

This patch is less invasive.

Bug: https://bugs.gentoo.org/761352
See also: https://github.com/mysql/mysql-server/pull/385
See also: https://bugs.mysql.com/bug.php?id=106034

### DIFF
--- a/libmysql/CMakeLists.txt
+++ b/libmysql/CMakeLists.txt
@@ -441,6 +441,19 @@ IF(HAS_WARN_FLAG)
     )
 ENDIF()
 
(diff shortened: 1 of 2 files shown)


thesamesam:

PR #506, opened 2023-11-15
On s390x, compile the FMA test with -O2 to avoid over-optimization

jonathan-albrecht-ibm:
The test for fused multiply add is compiled with -O3 by default but on s390x with g++ 11 or 12 at least, -O3 will optimize out the math instructions so the FMA optimization doesn't get a chance to be applied. Reduce the optimization level to -O2 so FMA can be detected.

This allows some of the mtr tests to pass that otherwise fail if FMA is applied: main.derived_limit, main.subquery_sj_mat_bka_nobnl and main.subquery_sj_all_bka_nobnl integration tests.


jonathan-albrecht-ibm:

PR #507, opened 2023-11-15
s390x is a big endian platform so use the big endian icu data dir

jonathan-albrecht-ibm:
This fixes the mtr tests:
  main.regular_expressions_utf-8
  main.regular_expressions_utf-8_icu_67


jonathan-albrecht-ibm:

PR #508, opened 2023-12-13
fix stale link about functions

PzaThief:
Just a tiny fix of comment.
I have signed OCA.


PzaThief:

PR #522, opened 2024-02-27
[RONDB-620] Foreign key column ordering mismatch

svenssonaxel:
`ndb_restore` fails to restore a foreign key with a different column order than the parent key.


svenssonaxel:

PR #523, opened 2024-03-01
fix typo change ENUM to SET

jiyfhust:
There seems to a typo when to check `m_elements.empty()`, because `m_elements` is related to `ENUM` and `SET`, the duplicate should be a typo problem.


jiyfhust:

PR #525, opened 2024-03-06
Fix operator name keywords causing obscure errors on -fno-operator-names

uoysip:
[Alternative operator name keywords](https://en.cppreference.com/w/cpp/language/operator_alternative) like `and`, `or`, `xor`, etc., are rarely used in MySQL and can cause obscure errors when the GCC flag `-fno-operator-names` is applied, for example:

```
/MySQL/storage/temptable/include/temptable/lock_free_type.h:125:75: error: template argument 2 is invalid
  125 |                                std::is_trivially_copyable<T>::value>::type> {
      |                                                                           ^
/MySQL/storage/temptable/include/temptable/lock_free_type.h:148:19: error: parse error in template argument list
  148 |     typename std::enable_if<std::is_same<T, long long>::value or
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  149 |                             std::is_same<T, unsigned long long>::value>::type> {
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Replace the few occurrences of operator name keywords to use their primary counter parts, like `&&`, `||`, and `!`.

### DIFF
--- a/sql/iterators/ref_row_iterators.cc
+++ b/sql/iterators/ref_row_iterators.cc
@@ -297,7 +297,7 @@ int PushedJoinRefIterator::Read() {
     if (error) {
       return HandleError(error);
     }
-  } else if (not m_is_unique) {
+  } else if (!m_is_unique) {
     int error = table()->file->ha_index_next_pushed(table()->record[0]);
     if (error) {
       return HandleError(error);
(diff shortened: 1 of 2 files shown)


uoysip:

PR #527, opened 2024-03-19
RONDB-615: Test case reproducing broken index after restore

svenssonaxel:
Reading table metadata while a unique index is disabled causes a
subsequent index rebuild to silently fail.

In mysql-test/suite/ndb/t/bug_rondb-615.test, the queries on lines 15,
23 and 32 should succeed, but instead produce "ERROR 1034 (HY000):
Incorrect key file for table 't1'; try to repair it".


svenssonaxel:

PR #528, opened 2024-03-20
Fix build failure with WITH_INNODB_EXTRA_DEBUG set

const-sambird:
Fixes two compile-time errors thrown with the WITH_INNODB_EXTRA_DEBUG flag set during compilation:
- in buf0lru.cc, a buf_block_t object incorrectly references `index` rather than through its member `ahi`
- in page0page.cc, zip debug checks are run despite the fact that the reference to the zipped page is no longer passed to the method


const-sambird:

PR #529, opened 2024-03-22
Fix OpenSSL linking when only static libs are available

samuel40791765:
https://github.com/mysql/mysql-server/commit/61b845549aa938d6c378f8920884f3f665dfc536 changed the way OpenSSL libraries are depended on.
We can use `WITH_SSL` to let mysql depend on custom OpenSSL locations, but the change is now dependent on Shared libraries being available. This wasn't the case before this commit, as the build would work fine with only Static libraries are available.

I am contributing on behalf of my employer Amazon Web Services.


samuel40791765:

PR #530, opened 2024-03-25
fix stale comment in my_command.h

nikezono:
A tinyfix of a comment to point [correct location](https://github.com/mysql/mysql-server/blob/trunk/sql/sql_parse.cc#L223-L259).


nikezono:
@mysql-oca-bot Thanks for the reply. I have signed OCA and created my user at bugs.mysql.com.

nikezono:

PR #536, opened 2024-04-12
fix stale comment in trx0undo.cc

nikezono:
TRX_UNDO_LOG_HDR_SIZE is no longer used/defined.
I have signed OCA.


nikezono:

PR #538, opened 2024-04-19
Fix build if UNIV_ZIP_DEBUG is enabled

0xAX:
If the `UNIV_ZIP_DEBUG` flag is enabled during build, the build process may fail with the error:

```
In function ‘bool buf_LRU_block_remove_hashed(buf_page_t*, bool, bool)’: error: ‘struct buf_block_t’ has no member named ‘index’
                                    ((buf_block_t *)bpage)->index));
                                                            ^~~~~
```

Because the 'index' field was moved from the buf_block_t structure to the buf_block_t.ahi_t in the d7114b14050 commit.

The second error appears in the `page_delete_rec()`. Since the `page_zip` parameter was removed from the function in the 85f9abf990750, the assert checks that are enabled in a case of `UNIV_ZIP_DEBUG` are failed. As there is no such parameter in the `page_delete_rec()` function anymore the both assertions are removed.

The commit fixes the both compilation errors.


0xAX:

PR #541, opened 2024-04-29
docs: multibyte collation for HandshakeResponse41

dveeden:
```python
import mysql.connector
c = mysql.connector.connect(
    host='127.0.0.1',
    port=3306,
    user="test",
    password="test",
    collation="utf8mb4_ja_0900_as_cs",
    ssl_disabled=True
)
cur = c.cursor()
cur.execute("SHOW SESSION VARIABLES LIKE '%collation%'");
for col in cur:
    print(col)
cur.close()
c.close()
```

Output:
```
('collation_connection', 'utf8mb4_ja_0900_as_cs')
('collation_database', 'utf8mb4_0900_ai_ci')
('collation_server', 'utf8mb4_ja_0900_as_cs')
('default_collation_for_utf8mb4', 'utf8mb4_0900_ai_ci')
```

This uses the `utf8mb4_ja_0900_as_cs` collation, which has ID=303. And 303 in hex is 0x012f. If only the lower 8 bits are sent this would be 0x2f (47 decimal). However the `0x01` is actually sent in the next byte.

image

### DIFF
--- a/sql-common/client.cc
+++ b/sql-common/client.cc
@@ -4186,10 +4186,10 @@ error:
   <tr><td>@ref a_protocol_type_int4 "int&lt;4&gt;"</td>
     <td>max_packet_size</td>
     <td>maximum packet size</td></tr>
-  <tr><td>@ref a_protocol_type_int1 "int&lt;1&gt;"</td>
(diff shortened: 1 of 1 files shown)


dveeden:

PR #543, opened 2024-05-18
Fix incorrect number of seconds in a day (86400, not 84600)

Pikamander2:
Hi there, I used a code search and found an incorrect number of seconds in a day (84600 instead of 86400) listed in one of your files. I recommend fixing it to avoid incorrect timings that could potentially lead to unexpected behavior.

https://github.com/mysql/mysql-server/blob/824e2b4064053f7daf17d7f3f84b7a3ed92e5fb4/sql/ssl_init_callback.cc#L219

Fix incorrect number of seconds in a day (86400, not 84600)


Pikamander2:

PR #504, opened 2023-11-10
Build failure with Protobuf 24.2 in Linux

gordonwwang:
**Question 1:**

In Linux, if you compile mysql-8.0.35 with protobuf 24.2, the error (Here is the full log:[build.log](https://build.stream.opencloudos.tech/kojifiles/work/tasks/4427/184427/build.log)):
```
/usr/bin/ld: /usr/lib64/libprotoc.so: undefined reference to symbol  '_ZN4absl12lts_2023080213base_internal12SpinLockWaitEPSt6atomicIjEiPKNS1_22SpinLockWaitTransitionENS1_14SchedulingModeE'
/usr/bin/ld: /usr/lib64/libabsl_spinlock_wait.so.2308.0.0: error adding symbols: DSO missing from command line.
```

According to the release notes of mysql-8.0.35, this issue should have been fixed in the current release, but this is not the case.

```
Building with WITH_PROTOBUF=system failed with Protobuf 22 or newer due to Protobuf 22 adding the Abseil dependency.  (Bug #111469, Bug #111623, Bug #35546459, Bug #35550389)
```

Reading the mysql source code, it seems that protobuf 22+ is only fixed on MacOS, so the problem still exists on Linux.

**Question 2:**

Run the FIND_OBJECT_DEPENDENCIES function to find dependent libraries. If the output of objdump is as follows:
```
NEEDED               libstdc++.so.6
NEEDED               libc.so.6
NEEDED ld-Linux-x86-64.2 
```

At this point, libc.so.6 can be processed normally and converted to "-lc"; 
However, "libstdc++.so.6" cannot be captured by the regular expression because it contains "++", so it cannot install the transposition "-lstdc++"; 
Although "ld-Linux-x86-64.2" can be captured by the regular expression, it is incorrectly converted to "-ld-Linux-x86-64", and the correct result should be "-ldl".

Function output error results are as follows:
```
-labsl_spinlock_wait -llibstdc -lgcc_s -lc -lld-linux-x86-64
```

I have to report this problem to mysql community, link: https://bugs.mysql.com/bug.php?id=113045

**How to reproduce:**

Environment: Linux x86-64 platform
Dependency: protobuf 24.2
Compile: Open -DWITH_PROTOBUF="system", then compile the mysql-8.0.35 code

**My fix:**
1. In cmake/protobuf.cmake, add fixes related to IF(LINUX AND WITH_PROTOBUF STREQUAL "system" AND PB_MINOR_VERSION VERSION_GREATER 21)
[...]

PR #544, opened 2024-05-24
Build failure with Protobuf 22+ in Linux

gordonwwang:
This is a re-opened PR (original PR：https://github.com/mysql/mysql-server/pull/504)

## **1. Problem:**

**Question 1:**
In Linux, if you compile mysql-8.0.35 with protobuf 24.2, the error (Here is the full log:[build.log](https://build.stream.opencloudos.tech/kojifiles/work/tasks/4427/184427/build.log)):
```
/usr/bin/ld: /usr/lib64/libprotoc.so: undefined reference to symbol  '_ZN4absl12lts_2023080213base_internal12SpinLockWaitEPSt6atomicIjEiPKNS1_22SpinLockWaitTransitionENS1_14SchedulingModeE'
/usr/bin/ld: /usr/lib64/libabsl_spinlock_wait.so.2308.0.0: error adding symbols: DSO missing from command line.
```

**Question 2:**

Run the FIND_OBJECT_DEPENDENCIES function to find dependent libraries. If the output of objdump is as follows:
```
NEEDED               libstdc++.so.6
NEEDED               libc.so.6
NEEDED               ld-linux-x86-64.so.2
```

At this point, libc.so.6 can be processed normally and converted to "-lc"; 
However, "libstdc++.so.6" cannot be captured by the regular expression because it contains "++", so it cannot install the transposition "-lstdc++"; 
Although "ld-Linux-x86-64.2" can be captured by the regular expression, it is incorrectly converted to "-ld-Linux-x86-64", and the correct result should be "-ldl".

Function output error results are as follows:
```
-labsl_spinlock_wait -llibstdc -lgcc_s -lc -lld-linux-x86-64
```
In fact, it should be:
```
-labsl_spinlock_wait -lstdc++ -lgcc_s -lc -ldl 
```

## **2. How to reproduce:**

Environment: Linux x86-64 platform
Dependency: protobuf 22+
Compile: Open -DWITH_PROTOBUF="system", then compile the mysql-8.0.35 code

## **3. Advice on this legacy question**

In 2023.10, the Release Note of mysql-8.0.35 claimed to have fixed the problem.
But I found that this issue still exists on Linux environments and seems to be fixed only on MacOS environments.
```
Building with WITH_PROTOBUF=system failed with Protobuf 22 or newer due to Protobuf 22 adding the Abseil dependency.  
```

In 2023.11, I proposed a fix on Linux, and PR was shut down by robots and not adopted.（PR link：https://github.com/mysql/mysql-server/pull/504）

It's 2024.05, and I find that **this problem persists even now. Neither mysql-8.0.36 nor 8.0.37 fixed this issue**.

As a Linux distribution operating system maintainer, I want mysql to be better and used by more people, so I am very concerned about this issue and hope mysql can fix this problem.


gordonwwang:

PR #545, opened 2024-06-02
Add "Traditional Chinese" collation for utf8mb4

danny50610:
Implement "Traditional Chinese" collation for utf8mb4 following the CLDR rule [reorder Hani Bopo] (using version v33.1), which sorts characters based on stroke count from smallest to largest.

Similar to the Chinese collation, we divides into 5 parts. However, for characters defined by CLDR, stroke order is used for sorting. Here, we use the 'short' version, as the original version contains approximately over 80,000 Chinese characters. There is uncertainty whether weights beyond 0xFFFF would function correctly and concerns about potential performance impacts.

Ref: https://github.com/mysql/mysql-server/commit/09f92de32649b6df5f9e9876df1e3c12aaef43ee
Ref: https://github.com/unicode-org/cldr/blob/release-33-1/common/collation/zh.xml#L5119


danny50610:

PR #547, opened 2024-06-06
Bug#114830	MySQL converts collation of date data type in ibd but data dictionary

venkatesh-prasad-v:
https://bugs.mysql.com/bug.php?id=114830

Problem
=======
When MySQL converts the charset of a table, it converts the date and time data types columns in data dictionary and SDI. However, the new collation_id in the sdi does not match that of the Field object.

Analysis
========
In MySQL, temporal types are always stored and compared using my_charset_latin1 charset.

During the execution of ALTER TABLE CONVERT TO CHARACTER SET, MySQL changes the charset and collation stored in data-dictionary and SDI for temporal columns to a different collation_id.  In practice, this new collation_id is ignored when such columns are stored/compared. Corresponding Field objects are not updated to use this new collation (Actually, Field objects for temporal types are hardcoded to use my_charset_latin1).

This new collation is not visible in I_S and SHOW CREATE TABLE output. It will be ignored by CREATE TABLE LIKE and rewritten by ALTER TABLE since both these statements use info from Field objects of source table to produce Create_field objects describing columns of new table/new version of the table.

For example:

CREATE TABLE a(dt DATETIME);
$ ./bin/ibd2sdi ./var/mysqld.1/data/test/a.ibd
"name": "dt",
"collation_id": 8

ALTER TABLE a CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci; $ ./bin/ibd2sdi ./var/mysqld.1/data/test/a.ibd
"name": "dt",
"collation_id": 224

ALTER TABLE a ENGINE = INNODB;
$ ./bin/ibd2sdi ./var/mysqld.1/data/test/a.ibd
"name": "dt",
"collation_id": 8

Particularly, for new temporal types that support fractional seconds collation_id from DD/SDI affects the InnoDB "precise type" describing the column and stored in .cfg. Because of this, incorrect collation_id in DD/SDI for such columns results in incorrect "precise type" in .cfg file.

As consequence we get Schema mismatch `(Column a precise type mismatch.)` error when ALTER TABLE IMPORT TABLESPACE compares the "precise type" from such .cfg file with "precise type" for column in table version being imported into. The latter is based on collation_id which ultimately comes from Field object and always corresponds to my_charset_latin1 for temporal types.

Note that this problem do not affect scenarios when we import table with .cfg file that was generated by MySQL's FLUSH TABLE FOR EXPORT command, as the latter doesn't use DD/SDI to calculate "precise type" but gets information from Field object in table being exported instead.

Solution
========
This commit changes ALTER TABLE CONVERT TO CHARACTER SET command to not alter the character set for temporal types stored in data-dictionary/SDI. In other words, we now force the server to always use my_charset_latin1 in DD/SDI for temporal types.

Approved by: Dmitry Lenev <dmitry.lenev@percona.com>

### DIFF
--- a/mysql-test/r/temporal_type_collation.result
+++ b/mysql-test/r/temporal_type_collation.result
@@ -1,0 +1,95 @@
+CREATE TABLE test (
+dt datetime primary key,
+datetxt varchar(10) GENERATED ALWAYS AS (DATE(dt)) STORED,
+timetxt varchar(10) GENERATED ALWAYS AS (TIME(dt)) STORED
+) ENGINE=InnoDB;
+SHOW CREATE TABLE test;
+Table	Create Table
+test	CREATE TABLE `test` (
+  `dt` datetime NOT NULL,
+  `datetxt` varchar(10) GENERATED ALWAYS AS (cast(`dt` as date)) STORED,
+  `timetxt` varchar(10) GENERATED ALWAYS AS (cast(`dt` as time)) STORED,
+  PRIMARY KEY (`dt`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
+FLUSH TABLE test FOR EXPORT;
+UNLOCK TABLES;
+#
+# Assert that the collation_id is my_charset_numericd in the SDI.
+#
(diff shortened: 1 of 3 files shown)


venkatesh-prasad-v:

PR #549, opened 2024-06-13
Added Azure Storage Select query support

0xar1:
Now Supports querying from Azure Table Storage.
Notes and readme will be added shortly.


0xar1:
Accidentally added this.

PR #550, opened 2024-06-15
Classic Protocol: Decoding MYSQL_TYPE_JSON in binary resultset

dveeden:
`MYSQL_TYPE_JSON` was missing on https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_binary_resultset.html


dveeden:

PR #556, opened 2024-07-07
bug#115484: Modify BuildRequires: to include gcc-toolset-12 for compat libraries on 9.0.0

sjmudd:
See: https://bugs.mysql.com/bug.php?id=115484

This is a minimal patch to fix the missing `BuildRequires:` for OL8/OL9, needed to build the `mysql-community-lib-compat` rpms.

While upstream RHEL8 / CentOS 8 is I believe EOL, it's not clear if OL8 is too. I believe longer support will be provided so building 9.0.0 and later may well still be handled on OL8/9.


sjmudd:

PR #558, opened 2024-07-15
Latency glitch fix

mronstro:


mronstro:
The patch is against MySQL 8.0, became mysql-trunk by mistake.

mronstro:
I confirm the code being submitted is offered under the terms of the OCA, and that I am authorized to contribute it."

PR #559, opened 2024-07-23
Reference COM_STMT_FETCH docs to itself instead of COM_STMT_CLOSE

Gbd199:


Gbd199:
hey, i'm not playing your agreement games, the docs have a few mistakes.

<https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_stmt_fetch.html> specifies "[0x19] COM_STMT_CLOSE" as its status instead of "[0x1C] COM_STMT_FETCH".

<https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_stmt_reset.html> specifies "[0x1A]" instead of "[0x1F]". (didn't include this in my change suggesitons)

do with this as you like

PR #560, opened 2024-07-24
A log record can span 3 log pages

mronstro:
The test case ndb.ndb_backup_nowait occasionally crashes in the REDO log execution
in the stepAhead function. The reason is that this test case can create PREPARE log
records that span 3 log pages (a log record can be 30000 + header + 4096 + 4 * 512 bytes
in the worst case). The parameter ZMIN_READ_BUFFER_LOG_SIZE is set to 2, thus only
log records spanning 2 log pages are safe to call stepAhead with. To ensure all log records
are safe to call stepAhead one needs to increase ZMIN_READ_BUFFER_LOG_SIZE to 3.

When this happens the node is unrecoverable until this patch is applied.


mronstro:

PR #561, opened 2024-07-29
Set $HOME for logrotate script

nisbet-hubbard:
> When executing from cron, the HOME term environment variable will be blank. Meaning: /usr/bin/mysqladmin won’t be able to find the file with the access credentials (user and password) and thus cannot execute the “flush-logs” command.

See: https://www.percona.com/blog/log-rotate-and-the-deleted-mysql-log-file-mystery/

This error is mostly invisible unless one looks at logrotate’s own log since the `mysqld.log` gets rotated even when `flush-logs` fails.

### DIFF
--- a/packaging/rpm-common/mysql.logrotate.in
+++ b/packaging/rpm-common/mysql.logrotate.in
@@ -30,7 +30,7 @@
 #       if test -x @bindir@/mysqladmin && \
 #          @bindir@/mysqladmin ping &>/dev/null
 #       then
-#          @bindir@/mysqladmin flush-logs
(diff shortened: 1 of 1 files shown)


nisbet-hubbard:

nisbet-hubbard:
I'm in communication with your open source team about a private sign-off. If that proves to be not possible, please feel free to fix this bug without me.

mysql-admin:
@nisbet-hubbard we are still not clear about the state of your OCA and who your in discussion with. 
You can email me with details at omer.barnir@oracle.com

Thanks
==Omer

PR #563, opened 2024-09-24
[RONDB-734] MTR fixes

svenssonaxel:
- Fix several regular expressions where a period matches any one character but a literal period was intended.
- Fix spelling.


svenssonaxel:

PR #566, opened 2024-10-07
Fix for resetting innodb_redo_log_capacity and innodb_buffer_pool_size when innodb_dedicated_server=1

mershadi:
Whenever innodb_dedicated_server is enabled, resetting innodb_buffer_pool_size to the 'default' value, resets it to the default of 128 MB instead of resetting it to the value calculated by innodb_dedicated_server.

We see a similar behavior with innodb_redo_log_capacity where it resets to the engine default of 100 MB when a custom value for innodb_redo_log_capacity was set in the my.cnf file at engine startup. However, if the value for innodb_redo_log_capacity was not set in the my.cnf file at engine startup, resetting the parameter sets it to the value calculated by innodb_dedicated_server i.e. nVCPU/2 GB.

The expected behavior is that the parameter value for innodb_buffer_pool_size and innodb_redo_log_capacity should reset to the value calculated by innodb_dedicated_server irrespective of weather of not it has been set in the my.cnf file.

This commit fixes bugs https://bugs.mysql.com/bug.php?id=115757 and https://bugs.mysql.com/bug.php?id=115759 .


mershadi:

mershadi:

PR #567, opened 2024-10-08
 bug#114904 Rename table with COPY does not change the constraint nam…

hotdb-cq:
…e, the bug reason is alter rename table by ALGORITHM COPY not update child table reference tablename in dict_foreign_t cache, The solution is reset chiild table reference table if refresh_fk is true in dd_table_check_for_child


hotdb-cq:

PR #568, opened 2024-10-08
Fix Incorrect Payload Length for COM_QUERY Example Packet

cybergarage:
This pull request addresses a bug in the MySQL packet examples where the payload length for COM_QUERY packets was calculated incorrectly. The specific fix involves 2B(+). This change ensures that the payload length is now computed accurately based on the actual length of the SQL query.


cybergarage:

PR #570, opened 2024-10-12
bug-114885/115353

hotdb-cq:
bug-114885/115353  Setting index_merge_union=off can result in an incorrect SQL execution plan (no matching row in const table) and incorrect results when none of the sub-parameters can be used for index merging.


hotdb-cq:

PR #572, opened 2024-10-12
bug-113302/113317 

hotdb-cq:
bug-113302/113317   When the index field = expression constant, and the index field is int/long, the result of the expression constant is decimal and there are valid decimals, the execution plan is determined to go through the index filtering, not Impossible WHERE, the result of the expression is forced to be converted into int/long to do index scanning to match, resulting in the result is wrong!


hotdb-cq:

PR #573, opened 2024-10-14
bug-110125 Unexpected query result fetched from an indexed column

hotdb-cq:
bug-110125 When the index field = function constant, and the index field is unsigned number, the result type of the function constant is string result, the execution plan is determined to go through the index filtering, not Impossible WHERE, the result of the function is forced to be converted into unsigned number to do index scanning to match, will be truncate to 0, resulting in the result is wrong


hotdb-cq:

PR #575, opened 2024-10-14
bug-115707 LOAD XML INFILE chooses child entity attribute instead of one within the tag

hotdb-cq:
bug-115707 Change the LOAD XML processor to choose the matched column names to elements by the depth that they are away from the tag from ROWS IDENTIFIED BY, such that those at the same level are chosen if present, then down through the nested XML.


hotdb-cq:

PR #583, opened 2024-12-20
Optimizing Permissions for CONNECTION_CONTROL_FAILED_LOGIN_ATTEMPTS Table

vidyadharchelluru:
The connection_control plugin introduces the CONNECTION_CONTROL_FAILED_LOGIN_ATTEMPTS
table, implemented through a plugin of the same name. This table tracks user login
attempts, storing information about the user, host, and the number of failed login
attempts. The primary purpose is to monitor failed login attempts, with the counter
resetting upon successful login. Unlike the host_cache table in the performance schema,
it doesn't contain additional sensitive information.

Currently, access to this table's contents requires superuser (SUPER) or CONNECTION_ADMIN
privileges. Without these, the table appears empty to users. However, given that
the PROCESS privilege already allows users to view information about other users,
including executed queries, it's more appropriate to use this privilege instead of
restricting access to SUPER users only. The CONNECTION_ADMIN privilege grants additional
capabilities that may be excessive for users who only need to monitor failed login attempts.

This patch modifies the relevant code to allow users with the PROCESS privilege
(in addition to those with CONNECTION_ADMIN) to view the contents of this table.
The SUPER privilege requirement has been removed from the list of necessary privileges.
This change is crucial to limit the privileges of monitoring users who perform checks
related to consecutive failed connection attempts, ensuring that access to this
important monitoring table is not overly restricted.

This commit fixes the bug https://bugs.mysql.com/bug.php?id=100743


vidyadharchelluru:
I confirm the code being submitted is offered under the terms of the OCA signed by Amazon, and that I am authorized to contribute it

PR #586, opened 2025-01-09
Report third decimal place for milliseconds

altmannmarcelo:
This patch adjust nice_timer to report third decimal place for
 milliseconds instead of only two as the current code does.

This is based on #84007 from @dveeden with the fixes applied.


dveeden (approved):

altmannmarcelo:

PR #589, opened 2025-01-14
BUG#117051: modify the delete want_priv by conditions

lihanxue:
https://bugs.mysql.com/bug.php?id=117051


lihanxue:
@mysql-oca-bot My application is still under review.

PR #593, opened 2025-01-31
Fix Example Packets for COM_STMT_PREPARE Response and Suggestion on Payload Specification

cybergarage:
Fixes for COM_STMT_PREPARE Response examples and a question with a proposed correction regarding the specification.

- [MySQL: COM_STMT_PREPARE Response](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_stmt_prepare.html)

# Suggestion on Payload Specification 
- The current payload specification reads the warning_count when packet_length > 12. However, both example packets have a packet_length of 12 (0x0C) and still include warning_count information. Should the specification be packet_length >= 12 instead, or is the condition unnecessary? 
- Note that MariaDB's payload specification does not have a conditional check on packet_length.
  - [COM_STMT_PREPARE response - MariaDB Knowledge Base](https://mariadb.com/kb/en/com_stmt_prepare/)
  
# Fix Example Packets 
- The first example contained extraneous characters (|), which have been removed.


daniyaalk:
@cybergarage I think those `|` characters have been added intentionally to illustrate the end of a MySQL Packet. Could probably be helpful if this was mentioned in a note.

cybergarage:
@mysql-oca-bot I confirm the code being submitted is offered under the terms of the OCA, and that I am authorized to contribute it.

PR #595, opened 2025-02-04
protocol_classic: clarify param type and flag

dveeden:
1. It is confusing to have `See enum_field_type` description for a 2 byte field as the type itself is only 1 byte.
2. Clearly describe that 128 / 0x80 / 0b10000000 is used as unsigned flag. Note that this is not the same as `UNSIGNED_FLAG` as that's 32 / 0x20 / 0b00100000


dveeden:

PR #598, opened 2025-02-11
docs: Use unversioned reference manual link

dveeden:
https://dev.mysql.com/doc/dev/mysql-server/9.2.0/ is for 9.2.0, but the reference manual link is for 8.0. This avoids this issue by using an unversioned link instead.


dveeden:

PR #599, opened 2025-02-18
bug #117500 query result error when empty result

liangximin2046:
fix a problem about scalar subquery, detail see
https://bugs.mysql.com/bug.php?id=117500


liangximin2046:

PR #600, opened 2025-02-19
Fix issues with mysql::serialization doc

dveeden:
- `varlen_integer_format` was mentioned twice in the list of what `field_data` can be.
- Doxygen didn't create an anchor for `#variable-length-integers` like other markdown software would do. So manually set the anchor.

With this PR:
image

Without this PR:
image


dveeden:

PR #601, opened 2025-02-19
Bug#111210 Extend pfs.replication_group_members to include MEMBER_WEIGHT

bgmilne:
In single-primary group-replication clusters, it may be necessary to
set group_replication_member_weight to ensure a host under (automated)
maintenance will not be elected to primary. Automation systems may also
need to identify the preferred secondary, which would be easier if the
value of group_replication_member_weight for each cluster member were
available in performance_schema.replication_group_members.


bgmilne:

PR #602, opened 2025-02-21
Return the correct collation for temporal results

dbussink:
When functions like `GREATEST`, `LEAST` and `COALESCE` return a temporal result, the collation is not set to `binary` as expected.

The collation for temporal types normally is always `binary` which is the case for fields and for functions inheriting from `Item_temporal_func`.

The comparison functions mentioned here though don't inherit from that and can return various types. Internally they do set the numeric collation correctly, but at the protocol level (and for what the `COLLATION` function returns), it is translated to the binary collation.

This translation does not happen before this change for the given functions, so we add it here.

Before:

```
mysql> select collation(date'2010-01-01');
+-----------------------------+
| collation(date'2010-01-01') |
+-----------------------------+
| binary                      |
+-----------------------------+
1 row in set (0.00 sec)

mysql> select collation(greatest(date'2010-01-01', date'2009-01-01'));
+---------------------------------------------------------+
| collation(greatest(date'2010-01-01', date'2009-01-01')) |
+---------------------------------------------------------+
| latin1_swedish_ci                                       |
+---------------------------------------------------------+
1 row in set (0.00 sec)
```

After:

```
mysql> select collation(date'2010-01-01');
+-----------------------------+
| collation(date'2010-01-01') |
+-----------------------------+
| binary                      |
+-----------------------------+
1 row in set (0.00 sec)

mysql> select collation(greatest(date'2010-01-01', date'2009-01-01'));
+---------------------------------------------------------+
| collation(greatest(date'2010-01-01', date'2009-01-01')) |
+---------------------------------------------------------+
| binary                                                  |
+---------------------------------------------------------+
1 row in set (0.00 sec)
```

This issue applies to all supported MySQL versions, starting from 8.0.x and later at least.

### DIFF
--- a/mysql-test/r/func_str.result
+++ b/mysql-test/r/func_str.result
@@ -673,6 +673,15 @@ latin2_general_ci	4
 select collation(replace(_latin2'abcd',_latin2'b',_latin2'B')), coercibility(replace(_latin2'abcd',_latin2'b',_latin2'B'));
 collation(replace(_latin2'abcd',_latin2'b',_latin2'B'))	coercibility(replace(_latin2'abcd',_latin2'b',_latin2'B'))
 latin2_general_ci	4
+select collation(greatest(date'2010-01-01', date'2011-01-01'));
+collation(greatest(date'2010-01-01', date'2011-01-01'))
+binary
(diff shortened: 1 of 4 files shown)


dbussink:

PR #606, opened 2025-04-07
Fix warning 1296

svenssonaxel:


svenssonaxel:

blaudden:
Thanks @svenssonaxel! It looks good. Most likely we will change to commit message to something like  "Fix temp table exist startup message" since we use 1296 code for almost everything.

PR #608, opened 2025-04-24
Fix incorrect byte size in mlog read_1_bytes comment

lee-jeong-geun:
Fix comment: 2 → 1 byte


lee-jeong-geun:

PR #611, opened 2025-05-13
Adding 'sb' instruction to spin_delay() for ARM v8.5 onward

salvatoredipietro:
We would like to propose this optimization for ARM architecture that, at runtime, it switches to SB instruction if supported by the system. SB (Speculation Barrier) is a modern barrier, available from armv8.5a. It achieves the same result as issuing ISB, but without having the CPU to drop its instruction pipeline.

Our performance evaluation based on 50 repetitions of the attached benchmark (inspired from https://bugs.mysql.com/bug.php?id=74832) conducted on Ubuntu 24.04 on m8g.24xlarge AWS instance using trunk MySQL branch, shows on average up to 28% reduction of the script execution time when `sb` instruction is used. Also linux perf shows ~11% drop of ut_delay function usage. Here the data:

```bash
sudo systemctl restart mysqld.service && sudo systemctl status mysqld.service

echo "Loading"
mysql -u root -pmysql < sql
sleep 5s

echo "Warmup"
for i in {1..5}; do 
    /usr/bin/time  --format="%e real,\t%U user,\t%S sys" ./psdoit 500
    sleep 5s;
done

echo "Benchmark"
for i in {1..50}; do 
    /usr/bin/time  --format="%e real,\t%U user,\t%S sys" --output=psdoit.sb.log --append ./psdoit 500
    sleep 5s;
done


# Results
$ cat psdoit.isb.log  | cut -f1 -d' ' | awk '{sum += $1; count++} END {print sum/count}'
32.445
$ cat psdoit.sb.log  | cut -f1 -d' ' | awk '{sum += $1; count++} END {print sum/count}'
24.504

# Perf ut_delay improvement
## ISB
+   18.87%    18.86%  connection  mysqld [.] ut_delay(unsigned long)

## SB
+    7.76%     7.76%  connection  mysqld  [.] ut_delay(unsigned long)
```

[sql.txt](https://github.com/user-attachments/files/20197312/sql.txt)
[psdoit.txt](https://github.com/user-attachments/files/20197313/psdoit.txt)

### DIFF
--- a/extra/gperftools/gperftools-2.15/src/base/spinlock.cc
+++ b/extra/gperftools/gperftools-2.15/src/base/spinlock.cc
@@ -37,6 +37,10 @@
 #include "base/spinlock_internal.h"
 #include "base/sysinfo.h"   /* for GetSystemCPUsCount() */
 
+#if defined(__GNUC__) && defined(__aarch64__)
+#include <sys/auxv.h>
+#endif // end __aarch64__
+
 // NOTE on the Lock-state values:
 //
 // kSpinLockFree represents the unlocked state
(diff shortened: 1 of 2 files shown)


salvatoredipietro:
Submitted OCA request

salvatoredipietro:

PR #612, opened 2025-05-19
add get_date and get_time for null_if func

liangximin2046:
fix bug for https://bugs.mysql.com/bug.php?id=118053


liangximin2046:

PR #613, opened 2025-05-20
fuzz: build all targets in the same output dir

catenacyber:
Allows simple script for oss-fuzz to pick them, see https://github.com/google/oss-fuzz/pull/13339


catenacyber:
Also, what do you think about https://github.com/google/oss-fuzz/tree/master/projects/mysql-server/targets ?

catenacyber:

PR #619, opened 2025-07-22
Optimize process of thread counter

wjunLu:
### Analysis
During MySQL's `sysbench point-select` performance testing, the server's CPU utilization approaches 100%. Using the perf top tool, it can be observed that the `dispatch_command` function in mysqld is a significant hotspot, accounting for over 20% of the CPU usage. Further analysis of `dispatch_command` reveals two atomic variable increment/decrement operations, each contributing to more than 40% of its execution time. These operations exhibit severe contention and represent clear bottlenecks, as illustrated in Figures 1 to 3.

Figure 1 - hotspots
image

Figure 2 -  The first bottleneck in `dispatch_command`
image

Figure 3 - The second bottleneck in  `dispatch_command`
image

### Optimization
The bottleneck code locations are as follows:
1. The first line performs an atomic increment operation on a shared variable.
2. The second line performs an atomic decrement operation on the same variable.

The primary reason these two lines become bottlenecks is due to severe contention caused by high-frequency atomic operations on the same variable.

Upon analyzing the related code, we found that these operations belong to the "running thread count" statistics and query module, which serves two key purposes:
- Increment/decrement the running thread count during SQL statement execution.
- Provide an interface to query the current running thread count.

The main workflow of this module is illustrated in Figure 4.

Figure 4 - Current main process of thread counter
thread1

The current running thread count statistics and query mechanism works by incrementing/decrementing the atomic counter during SQL statement execution, while queries simply returned the atomic variable's value. This design significantly impacted business performance.

The optimization removes the atomic operations during SQL execution and instead calculates the running thread count only during queries. This trade-off boosts business performance at the cost of slightly slower queries, as illustrated in the optimized workflow (Figure 5).

Figure 5 - The optimized process of thread counter 
thread2

### DIFF
--- a/sql/event_scheduler.cc
+++ b/sql/event_scheduler.cc
@@ -172,7 +172,6 @@ bool post_init_event_thread(THD *thd) {
 
   Global_THD_manager *thd_manager = Global_THD_manager::get_instance();
   thd_manager->add_thd(thd);
-  thd_manager->inc_thread_running();
   return false;
 }
 
@@ -192,7 +191,6 @@ void deinit_event_thread(THD *thd) {
   DBUG_PRINT("exit", ("Event thread finishing"));
   thd->release_resources();
   thd_manager->remove_thd(thd);
-  thd_manager->dec_thread_running();
   delete thd;
 }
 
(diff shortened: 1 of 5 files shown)


wjunLu:
My OCA has been approved, could anyone help review this PR?
image

wjunLu:

PR #622, opened 2025-07-26
Optimize NUMA Scheduling

wjunLu:
### Background
In MySQL OLTP scenarios, under high concurrency, the system by default schedules MySQL threads across all available CPU cores in the operating system, as shown in the left section of Figure 1. This leads to frequent cross-NUMA-node memory accesses by threads, which increases CPU overhead and consequently limits performance improvement.

Therefore, thread scheduling needs to be optimized to reduce cross-NUMA-node memory access latency, thereby improving overall system performance, as illustrated in the right section of Figure 1.

Figure 1. MySQL NUMA-Aware Scheduling Optimization Framework
image

### Optimization
The proposed optimized MySQL NUMA Scheduling enables fine-grained control over both foreground and background threads, thereby improving the efficiency of critical threads, reducing remote memory accesses, and ultimately enhancing overall system performance. (see Figure 2.)

Figure 2. Execution Workflow
schd

1. Background Threads
This feature targets 7 types of performance-critical background threads, primarily related to redo log and purge operations. These threads have a fixed count—only one instance per type—and start with the MySQL instance.
    - CPU Binding: Administrators can pin specific background threads to dedicated CPU core(s) via configuration.
    - Isolation & Efficiency: Proper parameter settings ensure CPU core isolation between different thread types, preventing scheduling contention and eliminating potential bottlenecks.

2. Foreground Threads
MySQL follows a thread-per-connection model, so foreground threads scale with client sessions. Similar to background threads:
    - NUMA-Aware Grouping: CPU cores are grouped by NUMA topology.
    - Strict Locality: During a session’s lifecycle, its thread migrates only within its assigned NUMA group, avoiding cross-node memory access penalties and maximizing data locality.
    - Dynamic Load Balancing: New sessions are assigned to the least-loaded NUMA group (measured by active session count per group), ensuring balanced resource utilization.


wjunLu:

PR #625, opened 2025-08-08
Replace ps|grep with pgrep for MySQL process detection

EvgeniyPatlan:
- Eliminates race condition where grep process counts itself
- More reliable process matching without shell pipeline artifacts  
- Cleaner code: single command vs pipeline of 3 commands
- Better performance: one process call instead of multiple


EvgeniyPatlan:

PR #633, opened 2025-10-20
Bug#119198 fix bug optimizer reverse index scan give incorrect result

srmppn:
# Description
This pull request fixes **Bug#119198**, which occurs when a reverse index range scan returns incorrect results. 

During SQL query optimization, the `used_key_parts` may need to be extended — for example, when a query requires ordering by both a column and its primary key. However, the optimizer did not properly check for this condition, causing `m_using_extended_key_parts` in the reverse_index_range_scan to remain false. As a result, the logic incorrectly falls into the `eqrange_all_keyparts` section instead, leading to wrong results:
```
if (eqrange_all_keyparts) {
      result = table()->file->ha_index_read_map(
          table()->record[0], last_range->max_key, last_range->max_keypart_map,
          HA_READ_KEY_EXACT);
 } 
```

# Proposed solution
Assign `used_key_parts` variable to retrieve possible extended key parts — similar to the approach used in `test_if_skip_sort_order` method. Additionally, I check for the maximum value using `std::max` to ensure we use extended key parts.

### DIFF
--- a/mysql-test/r/optimizer_reverse_index_range_scan_bug.result
+++ b/mysql-test/r/optimizer_reverse_index_range_scan_bug.result
@@ -1,0 +1,67 @@
+#
+# Bug#119198 Reverse index range scan give incorrect result.
+#
+CREATE TABLE temp_table (
+id bigint unsigned NOT NULL AUTO_INCREMENT,
+col1 tinyint unsigned NOT NULL,
+col2 tinyint unsigned NOT NULL,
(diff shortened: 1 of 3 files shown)


srmppn:
I have signed the OCA, and my application is currently under review.

srmppn:
the OCA application has beed approved.

srmppn:

PR #637, opened 2025-11-22
Bug #119442: Fix ORDER BY optimization doesn't recognize IS NULL as constant

akcube:
## What changes are proposed in this pull request?

**Why?**
This PR addresses [Bug#119442](https://bugs.mysql.com/bug.php?id=119442). The MySQL optimizer currently fails to treat columns filtered by `IS NULL` as "constants" during `ORDER BY` simplification. While an equality predicate (e.g., `col = 5`) allows the optimizer to skip sorting on that column, `col IS NULL` which functionally also guarantees a single value for sorting purposes is ignored.

This limitation forces the engine to perform unnecessary `filesort` operations even when a covering index exists that could satisfy the sort order if the `IS NULL` column were treated as a constant.

**What?**
This change updates the optimizer's `check_field_is_const()` logic to recognize `Item_func::ISNULL_FUNC` as a valid constant condition. By treating `IS NULL` as a constant, the optimizer can now rely on the index order for subsequent columns in the `ORDER BY` clause, eliminating `filesort` in these scenarios.

**How?**
1.  **Extended `check_field_is_const`**: Modified this function in `sql/sql_select.cc` to inspect `Item_func::ISNULL_FUNC` alongside the existing equality checks.
2.  **Sentinel Tracking**: Since `IS NULL` does not have a literal "value" like an integer or string, the `IS NULL` function pointer itself is stored as a sentinel in `const_item` to track state across recursive calls (e.g., inside `OR` conditions).
3.  **Conflict Resolution**: Logic to ensure logical consistency within `OR` branches:
    * `... AND (col IS NULL OR col IS NULL)` -> Recognized as **Constant**.
    * `... AND (col IS NULL OR col = 5)` -> Recognized as **Non-Constant** 
4.  **Helper Function**: Added `is_isnull_func()` to clean up some code.

## How is this tested?
**New MTR Tests (`mysql-test/t/order_by_isnull.test`)**
The suite covers the following scenarios:
* **Basic Covering Index**: Verified that `WHERE col IS NULL ORDER BY col, next_col` uses the index and removes `Using filesort`.
* **Compound Indices**: Verified that intermediate index columns filtered by `IS NULL` allow the index to provide ordering for subsequent columns.
* **OR Conditions**: Tested consistency checks to ensure `IS NULL OR IS NULL` is optimized, while `IS NULL OR col=1` is correctly treated as non-constant.
* **Ordering Semantics**: Verified that `NULL` sorting order (first in `ASC`, last in `DESC`) is preserved.
* **Optimizer Trace**: Verified via `INFORMATION_SCHEMA.OPTIMIZER_TRACE` that the column is now flagged with `"equals_constant_in_where": true`.

**Existing Regression Tests**
Updated `mysql-test/r/order_by_all.result` and `mysql-test/r/order_by_none.result`. These tests previously expected `Using filesort` for `IS NULL` queries; they have been updated to reflect the improved execution plan (`Using index`).

### DIFF
--- a/mysql-test/r/group_skip_scan.result
+++ b/mysql-test/r/group_skip_scan.result
@@ -4316,10 +4316,8 @@ INSERT INTO t1 VALUES (NULL, 1);
 INSERT INTO t1 VALUES (NULL, 2);
 EXPLAIN FORMAT=TREE SELECT MIN(f2) FROM t1 WHERE f1 IS NULL GROUP BY f1;
 EXPLAIN
--> Table scan on <temporary>  (rows=1)
-    -> Temporary table with deduplication  (rows=1)
-        -> Filter: (t1.f1 is null)  (rows=1)
-            -> Covering index skip scan for grouping on t1 using f1 over (f1 = NULL)  (rows=1)
+-> Filter: (t1.f1 is null)  (rows=1)
+    -> Covering index skip scan for grouping on t1 using f1 over (f1 = NULL)  (rows=1)
 
 SELECT MIN(f2) FROM t1 WHERE f1 IS NULL GROUP BY f1;
 MIN(f2)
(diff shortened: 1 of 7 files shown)


akcube:

PR #641, opened 2026-01-16
Bug#37397306 Fix partition pruning with non-constant defaults

FarihaIS:
## Description

This PR is a fix for a regression caused by the original fix of the bug #37397306, introduced in this commit: https://github.com/mysql/mysql-server/commit/c17d86b8f14b1a79fc1b95f0e3084c17a3aa62ad#diff-19fec1a1ff6223bd8530d4e61da2ffb72c2cb64d322481956e94bc3d5de2cd3aR480-R505 

The 8.0.42 changes in `set_used_partition()` only checks partition key fields explicitly provided in the `INSERT` column list. When a partition key field is omitted and uses a non-constant default (e.g. `CURRENT_TIMESTAMP`), this is not detected, causing incorrect partition selection and error 1748.

Fix this by counting partition key fields provided in the `INSERT` statement. If this count is lower than the total number of partition key fields, skip constant-based partition pruning.

Add test in `partition_pruning.test` in the main suite to test this fix.

## Release Notes

N/A

## How can this PR be tested?

Execute the `main.partition_pruning` test in mysql-test-run. This commit adds a test in this file to catch the regression.

### Before the fix

Running `main.partition_pruning` (with the added test) fails as below:
```
CURRENT_TEST: main.partition_pruning
mysqltest: At line 2656: Query 'CALL dp_insert_row(1)' failed.
ERROR 1748 (HY000): Found a row not matching the given partition set

The result from queries just before the failure was:
SET i = i + 1;
END WHILE;
SET partition_sql = CONCAT(partition_sql, '  PARTITION pmax VALUES LESS THAN MAXVALUE ENGINE = InnoDB\n);');
DROP TABLE IF EXISTS dp_tb_partition_issue_test;
# Create table
SET @sql = partition_sql;
PREPARE stmt FROM @sql;
EXECUTE stmt;
DEALLOCATE PREPARE stmt;
END|
CREATE PROCEDURE `dp_insert_row`(IN i BIGINT)
BEGIN
INSERT INTO `dp_tb_partition_issue_test` (`id`, `col1`) VALUES (i, 999999);
END|
CALL create_partition_table();
CALL dp_insert_row(1);
SELECT SLEEP(2);
SLEEP(2)
0
CALL dp_insert_row(1);
safe_process[8493]: Child process: 8494, exit: 1

 - the logfile can be found in '/quick-rebuilds/build-mysql/mysql-test/var/log/main.partition_pruning/partition_pruning.log'
```

### After the fix
Running `main.partition_pruning` (with the added test) succeeds:
```
[ 50%] main.partition_pruning                    [ pass ]  57844
```

## Basing the PR against the correct MariaDB version

## Copyright

### DIFF
--- a/mysql-test/r/partition_pruning.result
+++ b/mysql-test/r/partition_pruning.result
@@ -6440,3 +6440,60 @@ id	select_type	table	partitions	type	possible_keys	key	key_len	ref	rows	filtered
 Warnings:
 Note	1003	insert into `test`.`t1` values (20,rand())
 DROP TABLE t1;
+
+# Test for Bug#37397306 - partition pruning regression with CURRENT_TIMESTAMP default
+# This test verifies that INSERT with CURRENT_TIMESTAMP default values
+# works correctly with partitioned tables using RANGE partitioning on timestamp
+DROP DATABASE IF EXISTS test_partition_regression;
+CREATE DATABASE test_partition_regression;
+USE test_partition_regression;
+CREATE PROCEDURE create_partition_table()
+BEGIN
+DECLARE start_time DATETIME;
+DECLARE partition_sql TEXT;
+DECLARE i INT DEFAULT 0;
+# Start 10 seconds ago, create 41 partitions (covers ~40 seconds ahead)
+SET start_time = DATE_FORMAT(NOW() - INTERVAL 10 SECOND, '%Y-%m-%d %H:%i:%s');
+SET partition_sql = CONCAT(
+'CREATE TABLE `dp_tb_partition_issue_test` (\n',
+'  `id` bigint NOT NULL,\n',
+'  `col1` bigint NOT NULL,\n',
+'  `pt_timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,\n',
+'  PRIMARY KEY (`id`,`pt_timestamp`)\n',
+') ENGINE=InnoDB DEFAULT CHARSET=utf8mb3\n',
+'PARTITION BY RANGE (UNIX_TIMESTAMP(`pt_timestamp`))\n(\n'    );
+WHILE i < 41 DO
+SET partition_sql = CONCAT(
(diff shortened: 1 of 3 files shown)


FarihaIS:

mysql-admin:
Hi @FarihaIS,
Thanks for your contribution to fix this bug, however, this bug was already fixed in the April 2025 releases (8.0.42, 8.4.5, 9.3.0)
See:
- https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-42.html
- https://dev.mysql.com/doc/relnotes/mysql/8.4/en/news-8-4-5.html
- https://dev.mysql.com/doc/relnotes/mysql/9.3/en/news-9-3-0.html

Thanks
==Omer

FarihaIS:
Hi Omer, 

Thank you for reviewing this PR. I think there has been a misunderstanding though: to clarify, this change would fix a regression caused by the original fix of the bug #37397306, introduced in this commit: https://github.com/mysql/mysql-server/commit/c17d86b8f14b1a79fc1b95f0e3084c17a3aa62ad#diff-19fec1a1ff6223bd8530d4e61da2ffb72c2cb64d322481956e94bc3d5de2cd3aR480-R505 

The commit mentioned above introduced the regression where when a partition key field is omitted and uses a non-constant default (e.g. `CURRENT_TIMESTAMP`), this is not detected, causing incorrect partition selection and error 1748.

The 'How can this PR be tested' section in this PR adds a test that demonstrates this regression. Could you please review this PR again? I have added more details explaining that this PR is a fix for the regression introduced by the original fix of bug #37397306.

Please let me know if we can reopen this PR or if I need to open another PR, thank you.

mysql-admin:
Hi @FarihaIS,
Thanks for clarifying and apologies for the confusion.
Please see followup email (sent to your amazon mailbox on file in the MySQL bug system) 
Thanks
==Omer

PR #643, opened 2026-01-23
Bug #119784 Fix partition pruning with non-constant defaults regression

FarihaIS:
## Description

Fix regression caused by the original fix of the bug #37397306, introduced in this commit: https://github.com/mysql/mysql-server/commit/c17d86b8f14b1a79fc1b95f0e3084c17a3aa62ad#diff-19fec1a1ff6223bd8530d4e61da2ffb72c2cb64d322481956e94bc3d5de2cd3aR480-R505 

The original changes to `set_used_partition()` only check partition key fields explicitly provided in the `INSERT` column list. When a partition key field is omitted and uses a non-constant default (e.g. `CURRENT_TIMESTAMP`), this is not detected, causing incorrect partition selection and error 1748.

Fix this by counting partition key fields provided in the `INSERT` statement. If this count is lower than the total number of partition key fields, skip constant-based partition pruning.

Add test in `partition_pruning.test` in the main suite to test this regression and its fix.

## Release Notes

N/A

## How can this PR be tested?

Execute the `main.partition_pruning` test in mysql-test-run. This commit adds a test in this file to catch the regression.

### Before the fix

Running `main.partition_pruning` (with the added test) fails as below:
```
CURRENT_TEST: main.partition_pruning
mysqltest: At line 2656: Query 'CALL dp_insert_row(1)' failed.
ERROR 1748 (HY000): Found a row not matching the given partition set

The result from queries just before the failure was:
SET i = i + 1;
END WHILE;
SET partition_sql = CONCAT(partition_sql, '  PARTITION pmax VALUES LESS THAN MAXVALUE ENGINE = InnoDB\n);');
DROP TABLE IF EXISTS dp_tb_partition_issue_test;
# Create table
SET @sql = partition_sql;
PREPARE stmt FROM @sql;
EXECUTE stmt;
DEALLOCATE PREPARE stmt;
END|
CREATE PROCEDURE `dp_insert_row`(IN i BIGINT)
BEGIN
INSERT INTO `dp_tb_partition_issue_test` (`id`, `col1`) VALUES (i, 999999);
END|
CALL create_partition_table();
CALL dp_insert_row(1);
SELECT SLEEP(2);
SLEEP(2)
0
CALL dp_insert_row(1);
safe_process[8493]: Child process: 8494, exit: 1

 - the logfile can be found in '/quick-rebuilds/build-mysql/mysql-test/var/log/main.partition_pruning/partition_pruning.log'
```

### After the fix
Running `main.partition_pruning` (with the added test) succeeds:
```
[ 50%] main.partition_pruning                    [ pass ]  57844
```

## Copyright

### DIFF
--- a/mysql-test/r/partition_pruning.result
+++ b/mysql-test/r/partition_pruning.result
@@ -6440,3 +6440,60 @@ id	select_type	table	partitions	type	possible_keys	key	key_len	ref	rows	filtered
 Warnings:
 Note	1003	insert into `test`.`t1` values (20,rand())
 DROP TABLE t1;
+
+# Test for Bug#37397306 - partition pruning regression with CURRENT_TIMESTAMP default
+# This test verifies that INSERT with CURRENT_TIMESTAMP default values
+# works correctly with partitioned tables using RANGE partitioning on timestamp
+DROP DATABASE IF EXISTS test_partition_regression;
+CREATE DATABASE test_partition_regression;
+USE test_partition_regression;
+CREATE PROCEDURE create_partition_table()
+BEGIN
+DECLARE start_time DATETIME;
+DECLARE partition_sql TEXT;
+DECLARE i INT DEFAULT 0;
+# Start 10 seconds ago, create 41 partitions (covers ~40 seconds ahead)
+SET start_time = DATE_FORMAT(NOW() - INTERVAL 10 SECOND, '%Y-%m-%d %H:%i:%s');
+SET partition_sql = CONCAT(
+'CREATE TABLE `dp_tb_partition_issue_test` (\n',
+'  `id` bigint NOT NULL,\n',
+'  `col1` bigint NOT NULL,\n',
+'  `pt_timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,\n',
+'  PRIMARY KEY (`id`,`pt_timestamp`)\n',
+') ENGINE=InnoDB DEFAULT CHARSET=utf8mb3\n',
+'PARTITION BY RANGE (UNIX_TIMESTAMP(`pt_timestamp`))\n(\n'    );
+WHILE i < 41 DO
+SET partition_sql = CONCAT(
+partition_sql,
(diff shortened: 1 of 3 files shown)


FarihaIS:

jfg956:
Is there a verified bug for this regression on https://bugs.mysql.com/ ?

FarihaIS:
>Is there a verified bug for this regression on https://bugs.mysql.com/ ?

Not that I'm aware of, but I see above that an internal bug report was created at https://bugs.mysql.com/bug.php?id=119782

Do we need to create a non-internal bug report?

jfg956:
Bug#119782 is not an internal bug report.  It is the public bug which was opened to track this PR.

What I was asking is if there was a public bug report about the issue addressed by this PR.  The PR is well formed, with a test case, but to my knowledge, there has not been a discussion / validation about if this test is valid or not.  Without such discussion, I fear that this PR will not get the attention it deserves.

What I would suggest is to open a public bug report (on https://bugs.mysql.com/) about the observed behavior in a specific version (the test case of this PR would be a good start), explaining why this is unexpected, and suggesting an expected behavior.  If this gets validated as a bug, progress can be made on fixing it.

FarihaIS:
@jfg956 Thank you for the feedback! As per your suggestion, I have created a public bug report to explain this issue: https://bugs.mysql.com/bug.php?id=119784 and updated the PR title and git commit titles on my branch to reference this bug (this PR is not showing the latest versions of my commits as it is closed).

Please let me know if there is any further action item pending on my end before the PR can be reviewed. As well, should I create a new PR?

jfg956:
Thanks @FarihaIS for opening the bug.  The normal MySQL contribution workflow is not to review this PR, as you see it was closed by automation.  The contribution is tracked in below and will eventually be reviewed by Oracle and hopefully merged.
- https://bugs.mysql.com/bug.php?id=119782

(note that I am only a MySQL Community Member, I am not part of the bug processing team at Oracle, I am only trying to facilitate things here; and thanks for the contribution and opening the bug)

PR #644, opened 2026-01-27
Bug#119727: Fix wrong result for 'value BETWEEN a AND NULL' with temporal types

khaledalam:
## Description
- Fixes incorrect result when using BETWEEN with mixed temporal types (DATE, TIMESTAMP, YEAR) where one bound is NULL
- The expression `'2019-12-25' BETWEEN '2017-05-24' AND NULL` was returning `0` instead of `NULL`

## Root Cause
When YEAR is mixed with DATE/TIMESTAMP in BETWEEN, type aggregation returns `REAL_RESULT`. This bypassed temporal comparison, causing incompatible `val_real()` formats to be compared:
- DATE → `20191225.0`
- TIMESTAMP → `20170524000000.0`

Since `20191225.0 < 20170524000000.0`, the value appeared less than the lower bound, returning FALSE before checking the NULL upper bound.

## Solution
Added `all_temporal` check in `Item_func_between::resolve_type()` to force temporal comparison when all three arguments are temporal types, regardless of the aggregated `cmp_type`.

## Plan
- [x] Added new test file `func_between_temporal.test` with 8 test cases
- [x] Verified fix compiles without errors
- [x] All new tests pass

### DIFF
--- a/mysql-test/r/func_between_temporal.result
+++ b/mysql-test/r/func_between_temporal.result
@@ -1,0 +1,65 @@
+#
+# Bug#119727: wrong result for 'value between a and null' for certain
+#             temporal types
+#
+# Test case 1: DATE, TIMESTAMP, YEAR(NULL) - was returning 0, should be NULL
+CREATE TABLE t1 (value DATE, a TIMESTAMP, b YEAR);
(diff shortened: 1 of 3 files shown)


khaledalam:

PR #646, opened 2026-02-17
 Fix excessive generation of histogram background update errors 

tonychen2001:
## Description

When errors/warnings occur during background histogram updates, the conditions
in the diagnostic area are not cleared. This leads to future background
histogram updates emitting the same stale conditions to the error log which is
mis-leading at best.

We now clear the diagnostic area conditions in `prepare_session_context` to clear the
conditions from prior sessions.

This addresses Bug [#118999](https://bugs.mysql.com/bug.php?id=118999).

Additionally, there are paths for background histogram updates to still emit
errors. In particular, when an error occurs after the background_error_handler
is popped. This prevents certain errors during background histogram updates
from being throttled. Throttling of error messages regarding background histogram
updates was added in 1e39733b to prevent the error log from
being flooded with background histogram update warnings.

We now pop the error handler at a later stage in the background histogram
update to ensure errors are downgraded to warnings to ensure log throttling applies.

## How can this PR be tested?

The MTR `histogram_background_error` has been updated with test cases.

## Copyright


tonychen2001:

PR #654, opened 2026-03-30
Consolidate duplicated cmake flags in MySQL DEB packaging using share…

EvgeniyPatlan:
…d variable

  The debug and release build sections in rules.in and deb_debug.cmake
  contained two near-identical cmake invocations with the same flags
  repeated. Any cmake flag change required updating both places.

  Extract the common cmake flags into a CMAKE_COMMON_FLAGS Make variable
  in rules.in, referenced by both the release build block and the debug
  build block generated from deb_debug.cmake. Each build now contains
  only its unique flags (build type, compilation comment, PGO profile,
  maintainer mode).

  This also fixes two issues in the original debug cmake block:
  - MYSQL_MAINTAINER_MODE=0 was listed twice (lines 31 and 43)
  - INSTALL_DOCDIR was set explicitly but is unnecessary since INSTALL_LAYOUT=DEB already defines it


EvgeniyPatlan:
> Hi, thank you for your contribution. Please confirm this code is submitted under the terms of the OCA (Oracle's Contribution Agreement) you have previously signed by cutting and pasting the following text as a comment: "I confirm the code being submitted is offered under the terms of the OCA, and that I am authorized to contribute it." Thanks

PR #655, opened 2026-03-30
Fix stale sql_mode values in Query_event Doxygen documentation

altmannmarcelo:
The MODE_* hex values listed in the Q_SQL_MODE_CODE documentation in statement_events.h were outdated. The values from MODE_ANSI onward were all shifted by one bit compared to the actual definitions in sql/system_variables.h, and MODE_NO_ENGINE_SUBSTITUTION was missing entirely.

Updated all values to match system_variables.h, which is the authoritative source (sql_mode is written to the binlog directly from thd->variables.sql_mode without conversion).


altmannmarcelo:

PR #657, opened 2026-05-08
Fix "Row size too large" regression (Bug#120323)

keeper:
## Description                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                             
Fixes a regression introduced in 8.0.46 / 8.4.9 / 9.7.0 where CREATE TABLE fails with "Row size too large (> 8126)" for tables with many varchar(41-255) columns using `ROW_FORMAT=DYNAMIC`.                                                                         
                                                                                                                                                                                                                                                             
The fix Removes the `DATA_BIG_COL(col)` condition from the external-storage eligibility check in `get_field_max_size()` (storage/innobase/dict/dict0dict.cc), restoring the original row-size estimation behavior. 

Also updates `innodb_zip/bug52745.result` to remove a spurious warning that was a side effect of the same inflated estimation.                                                                                                                               
                                                                                                                                                                                                                                                             
The rationale is the `DATA_BIG_COL(col)` guard (added by Bug#39129182) prevents columns with col->len <= 255 from being counted as external-storage eligible during the DDL row-size check. This inflates the estimated row size for tables with many small-to-medium varchar columns, causing valid schemas that worked on all prior versions to be rejected.

Deadlock fix (Bug#39129182) is not affected: The deadlock fix relies on the `always_inlined` check for B-tree navigation columns. Our change is in the `!always_inlined` branch — a mutually exclusive code path that only affects non-navigation columns irrelevant to latch prediction.                                                                                                                                                                                                                            

Trade-off: Removing the check allows CREATE TABLE for schemas where a worst-case INSERT (every column at max length) could fail at runtime with DB_TOO_BIG_RECORD. This matches pre-8.0.46 behavior that has worked in production for years. The runtime error remains as a safety net.                                                                                                                                                                                                                     

## How can this PR be tested?                                                                                                                                                                                                                                                             
MTR test `bug120323_row_size_too_large_regression` verifies: 

1. 100 × varchar(200) latin1, DYNAMIC — the minimal repro that fails without the fix
2. 100 × varchar(255) latin1 — boundary case at the DATA_BIG_COL threshold (col->len = 255)
3. 100 × varchar(256) latin1 — control case that always worked (col->len > 255, DATA_BIG_COL = true)
4. 80 mixed varchar(41-255) columns — real-world schema with INSERT/SELECT and CREATE TABLE LIKE (upgrade path scenario)
5. 100 × varchar(40) latin1 — sanity check for columns at/below BTR_EXTERN_LOCAL_STORED_MAX_SIZE
 
Tests 1 and 2 are the regression cases — they fail without the fix and pass with it. Tests 3-5 confirm no regressions for schemas that were never affected.

MTR test `bug120323_row_size_insert_boundary` documents the known trade-off:

1. CREATE TABLE with 100 × varchar(200) succeeds (fix works)
2. INSERT with short data succeeds (normal usage)
3. INSERT with all columns at max length (200 bytes) fails with ER_TOO_BIG_ROWSIZE (expected — matches pre-8.0.46 behavior)

To verify the regression exists without the fix, revert the one-line change in dict0dict.cc (re-add DATA_BIG_COL(col) &&) and confirm tests 1 and 2 of the first test fail.

## Copyright

### DIFF
--- a/mysql-test/suite/innodb/r/bug120323_row_size_insert_boundary.result
+++ b/mysql-test/suite/innodb/r/bug120323_row_size_insert_boundary.result
@@ -1,0 +1,22 @@
+DROP DATABASE IF EXISTS bug120323;
+CREATE DATABASE bug120323;
+USE bug120323;
+#
+# Create a table with 100 x varchar(200) latin1 DYNAMIC.
+# With the fix, CREATE TABLE succeeds.
+#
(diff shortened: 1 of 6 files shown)


keeper:

polchawa-percona:
This fix addresses the most important part of the regression.

For the remaining minor part, I've filed a separate bug report:

https://bugs.mysql.com/bug.php?id=120504

PR #659, opened 2026-05-13
Fix ut_error() at ENABLE INNODB REDO_LOG when  innodb_checkpoint_disabled is set

nikezono:
### Summary

Fix a debug-build crash (`ut_error`) triggered by the following sequence:

```sql
ALTER INSTANCE DISABLE INNODB REDO_LOG;
SET GLOBAL innodb_checkpoint_disabled=1;
ALTER INSTANCE ENABLE INNODB REDO_LOG;
```

### Problem

`ALTER INSTANCE ENABLE INNODB REDO_LOG` calls `log_make_latest_checkpoint()` internally. When `srv_checkpoint_disabled` is set in debug builds, the inner helper method `log_request_checkpoint_validate()` returns `false`, but `log_request_latest_checkpoint()` treats that as an unrecoverable error and hits `ut_error`.

### Fix

Add an early-return guard in `log_request_latest_checkpoint()` (inside `#ifdef UNIV_DEBUG`) that returns `false` immediately when `srv_checkpoint_disabled` is set, before reaching the `ut_error` path.

### Test

Added redo_log_enable_checkpoint_disabled.test (debug-only) that reproduces the exact crash sequence and verifies that `ENABLE INNODB REDO_LOG` completes successfully.

### DIFF
--- a/mysql-test/suite/innodb/r/redo_log_enable_checkpoint_disabled.result
+++ b/mysql-test/suite/innodb/r/redo_log_enable_checkpoint_disabled.result
@@ -1,0 +1,25 @@
+#
+# Bug: ALTER INSTANCE ENABLE INNODB REDO_LOG hits ut_error in debug build
+#      when innodb_checkpoint_disabled is set.
+#
+# Repro steps:
+#   ALTER INSTANCE DISABLE INNODB REDO_LOG;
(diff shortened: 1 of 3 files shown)


nikezono:

PR #661, opened 2026-05-19
Log a warning when history list length exceeds threshold

tonychen2001:
This is a feature proposal.

## Description

Adds a throttled error-log warning when the **history list length** (HLL) exceeds a configurable threshold. A high HLL is an early indicator of a stuck purge or a long-running transaction that is preventing undo records from being reclaimed; left unaddressed it bloats undo space and can lead to degraded performance. Today this state is only observable by polling `information_schema.innodb_metrics`. Surfacing it in the error log makes the symptom discoverable to operators and log-aggregation pipelines without active monitoring.

Three new global system variables expose tuning and a kill switch (all dynamic):

| Variable | Default | Range | Purpose |
|---|---|---|---|
| `innodb_disable_hll_notification` | `OFF` | bool | Suppresses the warning entirely |
| `innodb_hll_notification_threshold` | `5,000,000` | `0..UINT32_MAX` | HLL value above which the warning is emitted |
| `innodb_hll_notification_interval_minutes` | `180` | `1..14400` | Minimum minutes between successive warnings |

## Release Notes

InnoDB now emits a warning to the MySQL error log when the history list length exceeds `innodb_hll_notification_threshold` (default 5,000,000). The warning is throttled to one message per `innodb_hll_notification_interval_minutes` window (default 3 hours) and can be disabled by setting `innodb_disable_hll_notification = ON`. The warning text is:

> The InnoDB history list length (HLL) is very high: *N*. Take actions to reduce the HLL, such as ending long-running transactions or tuning the purge parameters.

A high HLL typically indicates a long-running transaction blocking purge, or that purge is otherwise unable to keep up with undo generation. 

## How can this PR be tested?

Two new tests in the `innodb` suite:

```bash
./mysql-test-run.pl innodb.innodb_hll_notification
./mysql-test-run.pl innodb.innodb_hll_notification_throttle
```

`innodb_hll_notification` covers (1) knob range/clamping, (2) kill-switch suppression, (3) very-high threshold suppression, and (4) warning emission when the threshold is breached. `innodb_hll_notification_throttle` verifies that only one warning fires within an interval window even when the HLL stays elevated.

The tests use `START TRANSACTION WITH CONSISTENT SNAPSHOT` in a parallel session to pin a read view, blocking purge of undo records generated by a second session — a deterministic way to drive the HLL up.

## Copyright

### DIFF
--- a/mysql-test/r/all_persisted_variables.result
+++ b/mysql-test/r/all_persisted_variables.result
@@ -46,7 +46,7 @@ include/assert.inc [Expect 500+ variables in the table. Due to open Bugs, we are
 
 # Test SET PERSIST
 
-include/assert.inc [Expect 445 persisted variables in the table.]
+include/assert.inc [Expect 448 persisted variables in the table.]
 
 ************************************************************
 * 3. Restart server, it must preserve the persisted variable
@@ -54,9 +54,9 @@ include/assert.inc [Expect 445 persisted variables in the table.]
 ************************************************************
 # restart
 
-include/assert.inc [Expect 445 persisted variables in persisted_variables table.]
-include/assert.inc [Expect 445 persisted variables shown as PERSISTED in variables_info table.]
-include/assert.inc [Expect 445 persisted variables with matching peristed and global values.]
+include/assert.inc [Expect 448 persisted variables in persisted_variables table.]
+include/assert.inc [Expect 448 persisted variables shown as PERSISTED in variables_info table.]
+include/assert.inc [Expect 448 persisted variables with matching peristed and global values.]
 
 ************************************************************
 * 4. Test RESET PERSIST IF EXISTS. Verify persisted variable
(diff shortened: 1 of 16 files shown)


tonychen2001:

jfg956:
> innodb_disable_hll_notification	OFF	bool	Suppresses the warning entirely

I do not think this variable is needed: there is already [log_error_suppression_list](https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_log_error_suppression_list) for that.

> Surfacing it in the error log makes the symptom discoverable to operators and log-aggregation pipelines without active monitoring.

I am not sure aiming at making all miss-configuration (or degraded states) discoverable to operators in the error log is the right thing to do.  Also, having a log-aggregation pipeline without having active monitoring is not a very common situation (I see you are from AWS, and maybe this applies to RDS, but still, not super common).

IMHO, miss-configuration or degraded states should be discoverable via metrics (Statues, InnoDB Mectrics, or P_S).  Active monitoring might not be needed, but running a tool connecting to MySQL an giving recommendation is the right way to do this, not cluttering the logs.

To achieve above, we could keep `innodb_hll_notification_threshold` and adding a InnoDB Metrics indicating the threshold has be crossed.

PR #662, opened 2026-05-20
Bug#120480 Fix crash during relay log recovery

FarihaIS:
## Description

A replica crashes during relay log recovery when the relay log contains an `ALTER USER` or `GRANT` for a non-existent user. The crash is in `set_and_validate_user_attributes()` accessing `authentication_policy_list[0]` on an empty vector.

The root cause is a startup ordering issue in `mysqld.cc`: `ReplicaInitializer` starts replication threads before `update_authentication_policy()` populates `authentication_policy_list`. If the SQL thread applies an `ALTER USER` during this window, the out-of-bounds access crashes the server.

Fix by moving `update_authentication_policy()` before `ReplicaInitializer` in the startup sequence.

## Release Notes

N/A

## How can this PR be tested?

The bug is a race condition during server startup. To reproduce, apply the following patch to widen the race window:
```
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -8150,6 +8150,8 @@
   ReplicaInitializer replica_initializer(opt_initialize, opt_skip_replica_start,
                                          rpl_channel_filters,
                                          &opt_replica_skip_errors);
+
+  DBUG_EXECUTE_IF("delay_after_replica_init", { sleep(10); });
```

Then set up source-replica replication and run:
```
-- On source:
CREATE USER 'testuser'@'%' IDENTIFIED WITH mysql_native_password BY 'pass';
-- Wait for sync

-- On replica:
STOP REPLICA;
DROP USER 'testuser'@'%';

-- On source:
ALTER USER 'testuser'@'%' ACCOUNT LOCK;

-- Kill replica (kill -9), then restart with:
-- --relay-log-recovery --debug=+d,delay_after_replica_init
```

### Before this change:
The replica crashes during startup:
```
stl_vector.h:1263: Assertion '__n < this->size()' failed.
→ vector::operator[]
→ set_and_validate_user_attributes(...)
→ mysql_alter_user(...)
→ Query_log_event::do_apply_event(...)
Query: ALTER USER 'contactservice'@'%' ACCOUNT LOCK
```
### After this change:
The server starts up normally.

## Copyright

### DIFF
--- a/mysql-test/suite/rpl/r/rpl_relay_log_recovery_auth_policy_crash.result
+++ b/mysql-test/suite/rpl/r/rpl_relay_log_recovery_auth_policy_crash.result
@@ -1,0 +1,22 @@
+include/master-slave.inc
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the connection metadata repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START REPLICA; see the 'START REPLICA Syntax' in the MySQL Manual for more information.
+[connection master]
+call mtr.add_suppression("Replica SQL for channel.*");
+call mtr.add_suppression(".*coordinator and worker threads are stopped.*");
+CREATE USER 'testuser'@'%' IDENTIFIED WITH mysql_native_password BY 'pass';
+include/sync_slave_sql_with_master.inc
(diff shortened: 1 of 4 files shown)


FarihaIS:

FarihaIS:
Please see latest proposed changes here: https://github.com/FarihaIS/mysql-server/tree/bug120480, thank you

PR #664, opened 2026-06-04
Bug #44929 #92403 #110570: Fix tty password length handling

gitliuhc:
## Summary

This patch fixes the interactive TTY password input handling in the `mysql` client.

The previous implementation used a fixed-size buffer for interactive `-p` password input. As a result, passwords longer than 79 bytes could not be handled correctly. This also affects long token-style credentials used in token-based authentication scenarios.

When such long credentials are provided via interactive `-p` input, the input may be truncated or otherwise not processed correctly, which can lead to authentication failure and misleading error messages, for example:
ERROR 9125 (HY000): An error occurred while validating the access token. Please acquire a new token and retry.

One possible workaround is to pass the password directly using the --password option. However, this is not ideal from a security perspective because the password may be exposed through shell history or process command-line visibility. In addition, the mysql client explicitly shows the following warning in this case:
mysql: [Warning] Using a password on the command line interface can be insecure.

Therefore, this patch improves the interactive -p input path so that longer passwords and token-style credentials can be handled more safely, without requiring users to pass sensitive credentials directly on the command line.

## Related Bugs

* Bug #44929
* Bug #92403
* Bug #110570

## Changed File

* `sql-common/get_password.cc`

## Testing

Tested on:

* Windows 11
* Ubuntu 24.04

Test cases:

* 79-byte password
* 80-byte password
* 81-byte password
* 100-byte password
* 128-byte password
* 2391-byte token-style password

Both normal password-based authentication and long token-style credential input were tested.

## OCA
I have an approved Individual OCA.

## Review
I would appreciate your review and consideration of this patch. Please let me know if any changes are needed.

### DIFF
--- a/sql-common/get_password.cc
+++ b/sql-common/get_password.cc
@@ -73,10 +73,75 @@
 #define getpass(A) getpassphrase(A)
 #endif
 
+/**
+  Read a password from the terminal without echoing the typed characters.
+
+  Displays @p opt_message (or the default prompt "Enter password: ") and
+  reads characters one by one until the user presses Enter, Ctrl-C, or
+  until the internal buffer is exhausted.  The returned string is
+  allocated with my_strdup() (MY_FAE flag); the caller must release it
+  with my_free().
+
+  @param opt_message  Prompt to display before reading. Pass NULL to use
+                      the default "Enter password: ".
+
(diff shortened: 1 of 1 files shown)


gitliuhc:
I have an approved Individual OCA.
  image
image

gitliuhc:

PR #669, opened 2026-06-11
Use public variable name innodb_open_files in open-file timeout error

tonychen2001:
## Description

The error message ER_IB_MSG_TRYING_TO_OPEN_FILE_FOR_LONG_TIME referred to
"innobase_open_files", which is the internal C variable name in ha_innodb.cc,
not a name users can set. The variable is exposed to users as the
innodb_open_files system variable. Update the message to use the public name
so the suggested remediation is actionable, matching the neighboring
ER_IB_WARN_MANY_NON_LRU_FILES_OPENED message which already uses
innodb_open_files.

## Copyright


tonychen2001:

PR #673, opened 2026-06-17
Bug #120318 Fix stale partition statistics after ALTER TABLE REORGANIZE PARTITION

egAyush:
## Description

After `ALTER TABLE REORGANIZE PARTITION` on an InnoDB partitioned table, statistics in `mysql.innodb_table_stats` and `mysql.innodb_index_stats` are not recalculated, leaving every reorganized partition with `n_rows = 0` and zero index cardinality despite data being copied successfully. The stale statistics cause the optimizer to choose full table scans over appropriate indexes.

The root cause is an architectural divergence in the partition DDL path. Standard rebuilds (DROP COLUMN, OPTIMIZE TABLE) recalculate stats via `commit_inplace_alter_table_impl() → alter_stats_rebuild() → dict_stats_update()`. REORGANIZE PARTITION instead uses `ha_innopart::inplace_alter_partition()`, where after `copy_partitions()` completes no `dict_stats_update()` is invoked for the user table — the commit phase only updates internal DD tables, leaving the reorganized partitions with empty statistics.

Fix by adding a `dict_stats_update()` call after `copy_partitions()` in `ha_innopart::inplace_alter_partition()`, iterating modified partitions (skipping PART_NORMAL and PART_REORGED_DROPPED), using PERSISTENT or TRANSIENT recalc based on `innodb_stats_persistent`, and issuing a non-fatal warning on failure. Handles both regular and subpartitioned tables.

Add tests `innodb.partition_reorganize_stats` and `innodb.partition_reorganize_stats_transient` in the innodb suite.

## Release Notes
N/A

## How can this PR be tested?

Execute the following tests in `mysql-test-run`:
```
innodb.partition_reorganize_stats
innodb.partition_reorganize_stats_transient
```
- [x] This is a bug fix, and the PR is based against the earliest maintained branch in which the bug can be reproduced.

## Copyright

### DIFF
--- a/mysql-test/suite/innodb/include/partition_reorganize_stats.inc
+++ b/mysql-test/suite/innodb/include/partition_reorganize_stats.inc
@@ -1,0 +1,678 @@
+# Common test logic for partition reorganize statistics.
+# This file is sourced by both persistent and transient stats tests.
+#
+# Expected variables set by caller:
+#   $use_persistent_stats - 1 for persistent stats, 0 for transient stats
+#
+# Persistent stats verification: mysql.innodb_table_stats (n_rows column)
+# Transient stats verification: INFORMATION_SCHEMA.PARTITIONS (TABLE_ROWS column)
+
+--echo #
+--echo # Test 1: REORGANIZE PARTITION updates statistics correctly
+--echo #
+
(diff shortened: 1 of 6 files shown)


egAyush:

PR #680, opened 2026-06-18
Allow building mysqlclient without server deps

BlackDex:
Currently when I only want to build the mysqlclient library I'm forced to also provide the ncurses/readline and rpc, but the client does not need those.

With these changes will check if the `WITHOUT_SERVER` flag is set and skip those checks. Building the client still works this way.

I'm not sure if this might have other implications though with other parts of libraries able to build this way, but I think this should be fine.


BlackDex:
@mysql-oca-bot I confirm the code being submitted is offered under the terms of the OCA, and that I am authorized to contribute it.

akojima:
Hi,

Thanks for the patch. The direction makes sense: a client-only build should not need to discover dependencies used exclusively by server code.

However, I can’t accept this change as-is. `WITHOUT_SERVER` is not a library-only configuration: it still builds the interactive `mysql` CLI client, whose readline support is configured through the editline checks. Skipping those checks breaks that implicit contract, even if the build happens to complete in the proposed environment.

The RPC change looks appropriately scoped, but the editline part needs a more complete solution—either retain editline configuration for client programs, or introduce a genuinely library-only build configuration that excludes them.

PR #685, opened 2026-06-26
Bug#120706: Crash on SP reprepare of aggregate inside EXISTS subquery

vidyadharchelluru:
## Description

A stored procedure that caches a statement containing an aggregate inside an EXISTS subquery crashes with SIGSEGV in Item_sum::add_used_tables_for_aggr_func() whenever that cached statement is forced to reprepare between executions (for example, after changing a session variable that affects the statement's prepare-time metadata, such as group_concat_max_len).

### Root cause:
A forced reprepare makes validate_lex_and_execute_core() call free_lex() --
releasing the old item arena including the aggregate Item -- and re-parse the
expression. The new query block is parsed but never resolved: m_first_execution
is true, m_saved_base_items is empty and leaf_tables is nullptr. Despite that,
sp_lex_instr::execute_expression() still calls
Query_block::restore_cmd_properties(), which recurses into the unresolved
EXISTS subquery and runs update_used_tables() over a field list still
referencing the freed aggregate Item -- a use-after-free that faults when
Item_sum::add_used_tables_for_aggr_func() dereferences base_query_block.

### Solution:
Guard restore_cmd_properties()/bind_fields() with !m_first_execution in
sp_lex_instr::execute_expression(). When m_first_execution is true, exec_core()
has not run on this LEX, so nothing was saved by save_cmd_properties() and there
is nothing to restore. Skipping restore lets exec_core() resolve the reparsed
statement from scratch -- the same safe path taken on a genuine first execution.

## How can this PR be tested?

Added MTR test mysql mtr suite covering the reprepare path with exact reproducible steps mention in the bug report.

Execute the following tests in `mysql-test-run`:
```
bug120706
```

## Copyright

### DIFF
--- a/mysql-test/r/bug120706.result
+++ b/mysql-test/r/bug120706.result
@@ -1,0 +1,49 @@
+# Setup
+CREATE DATABASE IF NOT EXISTS bug_test;
+USE bug_test;
+DROP TABLE IF EXISTS test_table;
+Warnings:
+Note	1051	Unknown table 'bug_test.test_table'
+CREATE TABLE test_table (
+id       INT UNSIGNED NOT NULL AUTO_INCREMENT,
+value    VARCHAR(20) DEFAULT NULL,
+modified INT UNSIGNED NOT NULL,
+status   INT UNSIGNED NOT NULL,
+PRIMARY KEY (id),
+UNIQUE KEY (value)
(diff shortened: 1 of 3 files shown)


vidyadharchelluru:

PR #711, opened 2026-08-04
Rollback Prepared Transactions Asynchronously During Binlog Crash Recovery

SongLibing:
Problem:
========
Binlog recovery can block server startup for a long time when it must roll back a large prepared internal XA transaction. The rollback runs synchronously before the server becomes available.

Solution:
=========
Add an optional storage engine callback, recover_rollback_by_xid, for recovery-time rollback. During binlog recovery, InnoDB uses the callback to persistently convert a recovered prepared DML transaction back to ACTIVE state. The background recovery rollback thread then performs the expensive undo work asynchronously. DDL transactions still fall back to the normal synchronous rollback_by_xid path.

Add a binlog MTR test that covers prepared DML rollback, background rollback handoff, and repeated-crash safety before the rollback state change is flushed.

### What does this change do?

https://github.com/mysql/mysql-community/issues/100

### How was it tested?

### Contributor checklist

### AI assistance

If AI assistance was used, describe the tool(s) and extent of use: codex, only merge code from AliSQL to trunk.

### Areas touched
replication, innodb


henrikingo-oracle:
Thank you SongLibing for this.

We've uncovered that internally in Oracle this same issue and solution(s) have been discovered a couple years ago, discussed and documented, but not prioritized. We've made https://bugs.mysql.com/bug.php?id=114053 public so that it may benefit the review of this PR.

The review will take its time, we will probably want both a member from replication team as well as InnoDB team to look at this and this is obviously very critical part of the database, so they will take their time. Other than that everything seems to be in order now, oca signed and tests green.

design/server/100-async-rollback-during-binlog-recovery.md
```diff
@@ -0,0 +1,245 @@
...
+thread and waited for the rollback to complete. Undoing a large transaction
+row by row can therefore keep the server unavailable for hours.
+
+This change adds an optional recovery-specific storage-engine callback to
+perform a fast rollback handoff. InnoDB uses it to durably change the state of
+a prepared DML transaction back to active. The existing InnoDB recovery
+rollback thread then does the expensive row-by-row undo work in the background
+while server startup continues. The final transaction outcome does not change.
+
+## Functional Requirements
```
jujose-1, line 21: Thank you for documenting the requirements. Could you please consider expanding and regrouping this section? The current FR1 appears to describe a performance property, while NFR1 describes functional correctness.

 Would it make sense to cover the following requirement areas?

 Functional requirements
  - Transaction outcome and eligible transaction scope
  - Asynchronous handoff and synchronous fallback
  - Callback error handling and multi-engine transactions
  - Crash, restart, and shutdown behavior

 Non-functional requirements
  - Startup-time improvement and its measurable boundary
  - Replication, GTID, point-in-time recovery (PITR), and upgrade/downgrade compatibility
  - Resource usage and observability

design/server/100-async-rollback-during-binlog-recovery.md
```diff
@@ -0,0 +1,245 @@
...
+The design reuses existing undo-log states and redo operations. It adds no data
+dictionary objects, file-format fields, binlog events, or redo record types.
+Existing databases need no upgrade step, and older servers do not encounter a
+new on-disk representation.
+
+There is no change to external XA recovery: the new callback is selected for
+internal transactions by the existing binlog recovery decision path. There is
+also no change to security checks, SQL interfaces, replication protocols, or
+user-visible configuration. The interaction with the existing
+`innodb_force_recovery` setting is described above.
```
jujose-1, line 173: Could you please expand the impact assessment slightly beyond protocol compatibility?

It would be useful to confirm that GTID recovery, the recovered binlog position, and point-in-time recovery preserve the same final transaction outcome. It would also help to describe whether asynchronous replication or Group Replication can start while rollback is still running and what happens when an applier transaction conflicts with locks retained by the recovered transaction.

design/server/100-async-rollback-during-binlog-recovery.md
```diff
@@ -0,0 +1,245 @@
...
+
+The design reuses existing undo-log states and redo operations. It adds no data
+dictionary objects, file-format fields, binlog events, or redo record types.
+Existing databases need no upgrade step, and older servers do not encounter a
+new on-disk representation.
+
+There is no change to external XA recovery: the new callback is selected for
+internal transactions by the existing binlog recovery decision path. There is
+also no change to security checks, SQL interfaces, replication protocols, or
+user-visible configuration. The interaction with the existing
```
jujose-1, line 172: The design states that there is “no change to … user-visible configuration,” which appears consistent with the “Configuration / Knobs: No change” section in Issue #100.

However, the issue also refers to synchronous fallback “when the feature is disabled,” requests coverage for “disabled-feature behavior,” and shows a recovery path where the “option is ON.”

Could you please clarify the intended behavior and align the issue and design accordingly? If “enabled” refers only to whether a storage engine implements the optional callback, perhaps the option and disabled-feature wording in the issue could be updated. Otherwise, it would be helpful to describe the configuration mechanism in the design.

design/server/100-async-rollback-during-binlog-recovery.md
```diff
@@ -0,0 +1,245 @@
...
+- `sql/handler.h`: optional `handlerton` callback.
+- `sql/xa/recovery.cc`: callback selection for rollback decisions.
+- `storage/innobase/handler/ha_innodb.cc`: InnoDB registration and durable
+  handoff implementation.
+- `storage/innobase/trx/trx0roll.cc`: recovered-transaction handling and test
+  synchronization.
+- `storage/innobase/trx/trx0trx.cc`: scheduling tables for metadata-lock
+  acquisition before background rollback.
+
+## Alternatives Considered
```
jujose-1, line 220: Since the proposal adds a new storage-engine callback, could you please briefly explain why a recovery-specific callback is preferred over changing the behavior of the existing rollback_by_xid() callback?

It may also be useful to mention that adding a separate rollback worker was not selected because the existing InnoDB recovery rollback thread can already perform this work. A few sentences should be sufficient.

jujose-1 (commented):
Hi @SongLibing,

Thank you for the contribution. I reviewed the design from the replication and transaction-coordinator recovery perspective (the InnoDB-specific aspects would still need review from the InnoDB team).

The overall approach looks reasonable to me and appears to preserve the existing transaction outcome. I’ve left a few inline comments for clarification. Please take a look when you have a chance.

Thank you for working on this.

Regards,
Justin Jose

design/server/100-async-rollback-during-binlog-recovery.md
SongLibing: Correct me if I misunderstanding the definition of FR and NFR.  IMHO, FRs are in users perspective. It is the behaviors expected by users. And sure users can observe them. Before this feature, users observed binlog recovery blocking the server startup hours because rolling back a large transaction. For this feature, user expects binlog recovery should not block the server startup for a long time, rolling back a large transaction should delay startup by approximately the same amount as a small transaction. That is why I written it as the only FR.

About "Asynchronous handoff and synchronous fallback",
"Callback error handling and multi-engine transactions" etc., IMHO, they are things related to the design. If the design/solution is changed, they could be changed, but the requirements are not.

design/server/100-async-rollback-during-binlog-recovery.md
SongLibing: Thanks! In the original implementation in AliSQL, an variable was added to enable /disable this feature for cautious.  But we have not observed any problem in production for many years and It was merged into MariaDB without the variable. So I just removed the enable/disable description from the issue.

design/server/100-async-rollback-during-binlog-recovery.md
SongLibing: added

design/server/100-async-rollback-during-binlog-recovery.md
SongLibing: sure, added

SongLibing (commented):
Hi Justin,

Thanks for reviewing the design and given the comments.  The design is updated to address your comments. The PR part is unchanged, I explained why I didn't change it. Please check the detail in the comments.

Best Regards,
Libing Song

design/server/100-async-rollback-during-binlog-recovery.md
jujose-1: Thanks for the clarification and for updating the issue. I noticed that a few enable/disable references still remain in the proposed scope, pseudocode, and block diagram. Could you please update those as well so the issue consistently reflects that there is no configuration option?

design/server/100-async-rollback-during-binlog-recovery.md
jujose-1: Thanks for adding these details.

design/server/100-async-rollback-during-binlog-recovery.md
jujose-1: Thank you for the clarification. I agree that the implementation details can remain in the architecture and low-level design.

My remaining question is about the classification. The startup-time improvement appears to be a non-functional performance requirement, while preserving the transaction outcome appears to be a functional requirement. Could these be regrouped accordingly, with the performance expectation expressed in a testable way?

jujose-1 (commented):
Hi @SongLibing, 
Thank you for the clarification and discussion. I have replied to the review comments. Please take a look when you have a chance.

Regards,
Justin Jose

design/server/100-async-rollback-during-binlog-recovery.md
SongLibing: Thanks, updated others.

design/server/100-async-rollback-during-binlog-recovery.md
SongLibing: Thanks, exchanged FR1 and NFR1 and added performance expectation.

design/server/100-async-rollback-during-binlog-recovery.md
jujose-1: Thanks for the update. I can still see the following references in Issue #100:
  - “disabled-feature behavior” in Proposed Scope.
  - “option is ON” in the pseudocode.
  - “supported and enabled?” in the block diagram.

Could you please check whether those edits were saved?

SongLibing: Sorry, Justin. I forgot to save the edits, updated now.

design/server/100-async-rollback-during-binlog-recovery.md
```diff
@@ -0,0 +1,275 @@
+# Roll Back Prepared Transactions Asynchronously During Binlog Recovery
+
+Primary GitHub Issue: 100
```
henrikingo-oracle, line 3: That is actually now 'mysql/mysql-community#100' for historical reasons

Maybe it's clearer if you want to keep using your original mysql-community proposal (which makes sense of course) that we start moving them to mysql-server, so everything ends up in server repo?

SongLibing: Hi Henrik, 
I created a new issue in mysql-server repo and updated the design document.
https://github.com/mysql/mysql-server/issues/779

PR #714, opened 2026-08-06
Shrink NET::buff to reduce memory usage

SongLibing:
Problem:
========
NET::buff grows to accommodate large incoming packets but does not return to its initial size. Long-lived connections can therefore retain excess memory after processing occasional large requests.

Solution:
=========
Add the net_buffer_shrink_threshold session variable. When the current packet uses no more than half of an expanded buffer, count a shrink request. After the configured number of consecutive requests, halve the buffer without shrinking below net_buffer_length or IO_SIZE. Preserve the existing network error state if the allocation fails.

### What does this change do?
Optimize the memory usage of net::buff
https://github.com/mysql/mysql-community/issues/78

### Why is it needed?
It is needed to reduce memory usage and remain the server stable.
### How was it tested?

### Contributor checklist

### AI assistance

codex, merge the code to mysql-trunk

### Areas touched
NET, THD

### DIFF
--- a/mysql-test/t/shrink_net_buffer.test
+++ b/mysql-test/t/shrink_net_buffer.test
@@ -1,0 +1,173 @@
+# Test NET::buff shrinking for classic protocol connections.
+#
+# The test verifies that:
+#   1. A query larger than net_buffer_length grows NET::buff.
+#   2. Five consecutive small queries shrink the buffer by approximately half.
+#   3. Further small queries shrink it to, but never below, net_buffer_length.
+#   4. A later large query can grow the buffer again after it was shrunk.
+#
+# Expected result: PFS allocation statistics show grow, gradual shrink, a
+# stable lower bound, and successful regrowth in that order.
+
+--source include/have_nodebug.inc
+
+# =============================================================================
+# Prerequisites
+# =============================================================================
+
+if (`SELECT @@GLOBAL.performance_schema = FALSE`) {
+  skip Test requires @@GLOBAL.performance_schema = TRUE.;
+}
+
+let $instrument = `SELECT ENABLED FROM performance_schema.setup_instruments
+                    WHERE NAME = 'memory/sql/NET::buff'`;
+if ($instrument != YES) {
+  skip Test requires memory/sql/NET::buff instrumentation.;
+}
+
+# =============================================================================
+# Setup
+# =============================================================================
+
+let $old_net_buffer_shrink_threshold = query_get_value(
+  SHOW VARIABLES LIKE "net_buffer_shrink_threshold", Value, 1);
+SHOW VARIABLES LIKE 'net_buffer_length';
+
+# Generate padding for a query larger than net_buffer_length. The padding is
+# produced on the default connection so setup cannot affect con1's shrink count.
+let $query_padding = `SELECT REPEAT('x', 32768)`;
+
+CREATE USER 'shrink_test_user'@'localhost';
+SET GLOBAL net_buffer_shrink_threshold = 5;
+--connect(con1, localhost, shrink_test_user, , )
+let $con1_processlist_id = `SELECT CONNECTION_ID()`;
+--connection default
+--disable_query_log
+--eval SET @con1_processlist_id = $con1_processlist_id
+
+# Extra 39 bytes are allocated for
+# PSI header (32 B), net header (4 B), compression header (4 B)
+# They are not counted.
+
+let $pfs_net_buff_stmt =
+  SELECT COUNT_ALLOC - COUNT_FREE AS CURRENT_ALLOCATIONS,
+         CURRENT_NUMBER_OF_BYTES_USED - 39
+  FROM performance_schema.memory_summary_by_thread_by_event_name
+  WHERE EVENT_NAME = 'memory/sql/NET::buff'
+    AND THREAD_ID = (SELECT THREAD_ID
+                     FROM performance_schema.threads
+                     WHERE PROCESSLIST_ID = @con1_processlist_id);
+
+# =============================================================================
+# Test 1: A large incoming query grows NET::buff
+# =============================================================================
+
+--connection con1
+--echo [connection con1]
+--disable_result_log
+--eval SELECT 1 /* $query_padding */
+--enable_result_log
+
+--connection default
+--echo [connection default]
+--echo "NET::buff after a large query; capacity should have grown."
+--eval $pfs_net_buff_stmt
+
+# =============================================================================
+# Test 2: Five small queries shrink the grown buffer by about half
+# =============================================================================
+
+--connection con1
+--echo [connection con1]
+--disable_result_log
+SELECT 1;
+SELECT 1;
+SELECT 1;
+SELECT 1;
+--enable_result_log
+
+--connection default
+--echo [connection default]
+--echo "NET::buff after four small queries; capacity should not be shrunk."
+--eval $pfs_net_buff_stmt
+
+--connection con1
+--echo [connection con1]
+--disable_result_log
+SELECT 1;
+--enable_result_log
+
+--connection default
+--echo [connection default]
+--echo "NET::buff after five small queries; capacity should be about half."
+--eval $pfs_net_buff_stmt
+
+# =============================================================================
+# Test 3: Five more small queries reach net_buffer_length
+# =============================================================================
+
+--connection con1
+--echo [connection con1]
+--disable_result_log
+SELECT 1;
+SELECT 1;
+SELECT 1;
+SELECT 1;
+SELECT 1;
+--enable_result_log
+
+--connection default
+--echo [connection default]
+--echo "NET::buff after ten small queries; capacity should reach its minimum."
+--eval $pfs_net_buff_stmt
+
+let $net_buffer_length = query_get_value(
+  SHOW VARIABLES LIKE 'net_buffer_length', Value, 1);
+let $current_buffer_capacity = `$pfs_net_buff_stmt`;
+
+if ($current_buffer_capacity == $net_buffer_length) {
+  --echo NET::buff capacity is equal to net_buffer_length.
+}
+
+# =============================================================================
+# Test 4: Further small queries do not shrink below net_buffer_length
+# =============================================================================
+
+--connection con1
+--echo [connection con1]
+--disable_result_log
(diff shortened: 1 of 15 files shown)


mysql-test/t/shrink_net_buffer.test
```diff
@@ -0,0 +1,184 @@
...
+--echo "NET::buff after a large query; capacity should have grown."
+SELECT COUNT_ALLOC - COUNT_FREE AS CURRENT_ALLOCATIONS,
+       CURRENT_NUMBER_OF_BYTES_USED
+FROM performance_schema.memory_summary_by_thread_by_event_name
+WHERE EVENT_NAME = 'memory/sql/NET::buff'
+  AND THREAD_ID = (SELECT THREAD_ID
+                   FROM performance_schema.threads
+                   WHERE PROCESSLIST_ID = @con1_processlist_id);
+
+# =============================================================================
```
phulakun, line 76: Kindly add test to verify behavior with small queries less than net_buffer_shrink_threshold.

sql/sql_class.cc
```diff
@@ -584,6 +584,61 @@ THD::Attachable_trx_rw::Attachable_trx_rw(THD *thd)
...
+
+  /*
+    Store original status to prevent failure. Only last_error and error
+    need to be stored.
+  */
+  uint old_last_errno = net->last_errno;
+  uchar old_error = net->error;
+
+  if (net_realloc(net, target_buffer_capacity)) {
+    net->last_errno = old_last_errno;
```
phulakun: If net_realloc() hits OOM: 
   my_realloc(..., MYF(MY_WME)) reports EE_OUTOFMEMORY through my_error().
This can populate the current THD diagnostics area.

Restoring net->error and net->last_errno repairs only the NET state. The THD diagnostics area can still hold OOM.

Perhaps shrink should call my_realloc(..., MYF(0)) instead. Something like,

// sql-common/net_serv.cc
// Note: net_shrink is a new function.
bool net_shrink(NET *net, size_t length) {
  assert(net->write_pos == net->buff);
  assert(length < net->max_packet);

  const size_t target =
      (length + IO_SIZE - 1) & ~(static_cast<size_t>(IO_SIZE) - 1);

  uchar *buff = static_cast<uchar *>(my_realloc(
      key_memory_NET_buff, net->buff,
      target + NET_HEADER_SIZE + COMP_HEADER_SIZE, MYF(0)));

  if (buff == nullptr) return true;

  net->buff = buff;
  net->buff_end = buff + target;
  net->write_pos = buff;
  net->read_pos = buff;
  net->max_packet = static_cast<ulong>(target);
  return false;
}

What do you think?

sql/sys_vars.cc
```diff
@@ -3086,6 +3086,18 @@ static Sys_var_ulong Sys_net_buffer_length(
...
     NO_MUTEX_GUARD, NOT_IN_BINLOG, ON_CHECK(check_net_buffer_length));
 
+#define DEFAULT_NET_SHRINK_COUNT_THRESHOLD 5
+
+static Sys_var_ulong Sys_net_buffer_shrink_threshold(
+    "net_buffer_shrink_threshold",
+    "If required size of NET::buff currently is less than half the allocated "
+    "one, shrink will be requested. When number of consecutive requests is "
+    "equal or greater than this value, shrink will be triggered. 0 means "
+    "never do shrink.",
```
phulakun: Perhaps its good to mention that NET::buff is not shrinked below net_buffer_length.

phulakun (commented):
Hello SongLibing!

Thank you for working on this. It looks good. Please find few comments.

Regards,
Praveen

sql/sql_class.cc
SongLibing: Added net_shrink function into net_srv.cc, net_shrink doesn't report any error.

sql/sys_vars.cc
SongLibing: sure, added.

mysql-test/t/shrink_net_buffer.test
SongLibing: For test2 and test3, I added query to show the allocated buffer, just after the forth SELECT 1.  the allocated buffer should no change.

SongLibing:
> Hello SongLibing!
> 
> Thank you for working on this. It looks good. Please find few comments.
> 
> Regards, Praveen

Hi Praveen,

Thanks for your review comments. I addressed all problems you mentioned. Please have a look.

Regards, Libing

phulakun (commented):
Hello Libing!

Thank you for addressing review comments.
Changes looks good to me. Kindly wait for review from Harin.

Regards,
Praveen

include/mysql_com.h
```diff
@@ -946,6 +946,8 @@ typedef struct NET {
     to maintain the server internal instrumentation for the connection.
   */
   void *extension;
+  /** Number of consecutive requests to shrink NET::buff. */
+  unsigned int shrink_requests;
```
harinvadodaria: Adding a new variable in NET would change sizeof(NET) and in turn sizeof(MYSQL) because NET is part of MYSQL structure. This is also captured by changes in mysql.h.pp. Such a change would need a bump in libmysql version.

Instead, please extend structure NET_SERVER in mysql_com_server.h and initialise in init_net_server_extension().

This way, ABI changes are avoided and no version bump is needed.

sql/sys_vars.cc
```diff
@@ -3086,6 +3086,18 @@ static Sys_var_ulong Sys_net_buffer_length(
     VALID_RANGE(1024, 1024 * 1024), DEFAULT(16384), BLOCK_SIZE(1024),
     NO_MUTEX_GUARD, NOT_IN_BINLOG, ON_CHECK(check_net_buffer_length));
 
+#define DEFAULT_NET_BUFFER_SHRINK_THRESHOLD 5
```
harinvadodaria, line 3089: What is the rationale behind this default value? While I understand that shrinking a buffer reduces memory usage. It also means that every now and then, perf penalty is paid to realloc required buffer. Can you share some performance data based on your observations?

include/mysql_com.h
```diff
@@ -1096,6 +1098,7 @@ void net_end(struct NET *net);
 void net_clear(struct NET *net, bool check_buffer);
 void net_claim_memory_ownership(struct NET *net, bool claim);
 bool net_realloc(struct NET *net, size_t length);
+bool net_shrink(struct NET *net, size_t length);
```
harinvadodaria: Once `shrink_request` is moved out of NET, this function can be removed from mysql_com.h.
Instead, it can be part of server code itself.

sql/sql_parse.cc
```diff
@@ -2530,7 +2530,17 @@ bool dispatch_command(THD *thd, const COM_DATA *com_data,
...
     thd->mem_root->Clear();
 
-    /* SHOW PROFILE instrumentation, end */
+  if (thd->is_classic_protocol()) {
+    Protocol_classic *protocol = thd->get_protocol_classic();
+    NET *net = protocol->get_net();
+    if (net->max_packet > thd->variables.net_buffer_length &&
+        net->max_packet > IO_SIZE &&
+        thd->variables.net_buffer_shrink_threshold != 0) {
+      thd->try_shrink_net_buffer(net, protocol->get_packet_length());
```
harinvadodaria: All of the above mentioned checks should be folded into try_shrink_net_buffer(). You already have assertions for same conditions in the function.

sql-common/net_serv.cc
```diff
@@ -173,6 +173,7 @@ bool my_net_init(NET *net, Vio *vio) {
   net->last_errno = 0;
 #ifdef MYSQL_SERVER
   net->extension = nullptr;
+  net->shrink_requests = 0;
```
harinvadodaria: Can be removed once shrink_requests is moved as suggested in one of the previous comments.

sql-common/net_serv.cc
```diff
@@ -258,6 +259,34 @@ bool net_realloc(NET *net, size_t length) {
...
+  Shrink net buffer.
+
+  @param net      NET handler
+  @param length   shrink net buffer to length. length should smaller
+                  than the original buffer size
+
+  @retval  true   failed to shrink
+  @retval  false  Shrunk to length successfully
+*/
+bool net_shrink(NET *net, size_t length) {
```
harinvadodaria, line 272: Please fold it into try_shrink_net_buffer() or define a separate function within server layer itself.

harinvadodaria (commented):
Hi Libing,

Thanks for submitting the patch.
Please see some comments below

sql/sql_parse.cc
SongLibing: It is originally in the try function. I moved the check out try_shrink_net_buffer, because I think it will avoid a function call in most of the situation. So it is faster.

include/mysql_com.h
SongLibing: Got it, thanks! moved it into NET_SERVER.

include/mysql_com.h
SongLibing: Moved it into mysql_com_server.h

sql/sql_parse.cc
SongLibing: Folded into try_shrink_net_buffer(). The function is only called once in the query life, it is fine.

sql-common/net_serv.cc
SongLibing: removed

sql-common/net_serv.cc
SongLibing: Added #ifdef MYSQL_SERVER to make it server only.

sql/sys_vars.cc
SongLibing: The target is to reduce memory usage for very large net buffer.  The feature is enabled in production by default, we use a relative mild config value 5 with cautious. Because we don't want the config impact the instances in which queries are often large than  net_buffer_length. And the value 5 works fine in production. Below is an real case we observed in production : you can see the memory usage is much less and the instance stays stable after the server is updated to the version with this feature.
image

SongLibing:
Hi @harinvadodaria ,
Thanks for the great suggestions. I addressed all problems. Please have a look.

harinvadodaria (approved):
Thank you for the updated patch, @SongLibing .
Approved.

PR #723, opened 2026-08-19
Bug#121124 Ignore unchanged CHECK columns for online DDL

modasserbillah:
### What does this change do?

Only compare a CHECK-referenced column's type when its `Create_field` represents a column changed by the ALTER. It also adds MTR coverage for both INSTANT and INPLACE ENUM extensions when an unrelated enforced CHECK references `DATETIME(6)`.

### Why is it needed?

When any `MODIFY COLUMN` sets `ALTER_CHANGE_COLUMN`, the CHECK-constraint evaluation code currently examines unchanged columns too. `Item_field` reports an unchanged `DATETIME(6)` as `MYSQL_TYPE_DATETIME`, while `Create_field` reports it as `MYSQL_TYPE_DATETIME2`. MySQL therefore falsely concludes that the checked column changed and requires a table copy for an otherwise metadata-only ENUM extension.

Fixes [MySQL Bug #121124](https://bugs.mysql.com/bug.php?id=121124).

### How was it tested?

- [ ] Ran the relevant full suite (name it): N/A

`git diff --check` passes and the changed C++ code was formatted with the repository configuration. A local Debug build was attempted on macOS, but the host's global Homebrew Protobuf 29.1 headers conflict with MySQL's bundled Protobuf 24.4 before the build reaches the changed code. The upstream Linux CI build and MTR run are therefore the authoritative validation.

The initial GitHub build, MTR, and format jobs also fail before checkout because the repository's `pull_request_target` workflows do not opt in to checking out fork PR code under the current `actions/checkout` security policy. No PR code is compiled, formatted, or tested in those failed jobs.

### Contributor checklist

- [x] I have signed the OCA with the email on these commits

### AI assistance

OpenAI Codex assisted with root-cause analysis, the code change, regression-test preparation, and this pull request description.

### Areas touched

- SQL-layer ALTER TABLE CHECK-constraint evaluation
- `check_constraints` MTR coverage


modasserbillah:
I have signed an OCA which is under review. The CI workflows seem to block the checkout step for PRs from forks. @gopshank please let me know if I need to follow a different procedure for filing the PR.

gopshank:
> I have signed an OCA which is under review. The CI workflows seem to block the checkout step for PRs from forks. @gopshank please let me know if I need to follow a different procedure for filing the PR.


Sure, we shall check and get back.

RidhaOracle:
@modasserbillah Can you rebase your branch on latest trunk to pick up the fixed Github CI/CD actions ?

modasserbillah:
@RidhaOracle done!

modasserbillah:
The formatting failure lines reported are not part of my PR.

modasserbillah:
Looks like the failures are unrelated @RidhaOracle please let me know if there's something to do on my side.

modasserbillah:
@RidhaOracle @phulakun just following up in case you want me to make any changes.

sql/sql_table.cc
```diff
@@ -20791,7 +20791,8 @@ static bool is_any_check_constraints_evaluation_required(
             continue;
 
           // Check if data type is changed.
-          if (!my_strcasecmp(system_charset_info, itm_fld->field_name,
+          if (fld.change &&
```
phulakun, line 20919: @modasserbillah Analysis on the issue is correct. 

Internal type representation for DATATIME type field is:
    item_fld->data_type() = MYSQL_TYPE_DATETIME and
    fld.sql_type == MYSQL_TYPE_DATETIME2.
Because of which ALTER operation is disallowed.

Thank you for providing fix for it.  But this fix may not work in all the cases. For instance,

mysql> CREATE TABLE t_datetime (
    ->   id BIGINT UNSIGNED NOT NULL,
    ->   view_type ENUM('A','B') NOT NULL,
    ->   checked_at DATETIME(6) DEFAULT NULL,
    ->   PRIMARY KEY (id),
    ->   CONSTRAINT chk_checked_at CHECK (checked_at IS NULL)
    -> ) ENGINE=InnoDB;

ALTER TABLE t_datetime MODIFY COLUMN checked_at datetime(6) DEFAULT NULL;

In this case, all attributes of `checked_at` column remain as is but `fld.change` is a non-null pointer value.

Better solution would be to use the right type to compare here. 

How about changing condition to

          // Check if data type is changed.
          if (!my_strcasecmp(system_charset_info, itm_fld->field_name,
                             fld.field_name) &&
              (itm_fld->data_type() != real_type_to_type(fld.sql_type)))

?

What do you think?

modasserbillah: Thanks @phulakun, I think it's a great idea and addresses the root cause. I didn't know about `real_type_to_type`! 

I think we should also keep the `fld.change` condition to avoid going into the branch if the check related columns were not changed. I have pushed an update to keep both and also added relevant tests. Thanks again!

phulakun (approved):
Hello @modasserbillah 

Thank you for addressing the review comments. Changes look good to me.

Regards,
Praveen

modasserbillah:
Thanks @phulakun , what's the next step to get this merged? Also, is there a way to ship this to version 8.0.42?

modasserbillah:
@gopshank @RidhaOracle please guide me to the next steps to get this merged, thanks!

RidhaOracle:
Triggering the CI/CD pipelines and then will move it to integrate.

modasserbillah:
Thanks @RidhaOracle , as noted before, the formatting errors are unrelated to this change.

phulakun:
> Thanks @phulakun , what's the next step to get this merged? Also, is there a way to ship this to version 8.0.42?

Generally non-security patches are **not** shiped with the LTS releases. Can you please share justification for shiping it with LTS releases(8.4+)? Please note that, no non-security patches are allowed in 8.0 as it is in EOL phase.

modasserbillah:
> Can you please share justification for shiping it with LTS releases(8.4+)?

It's just that my application is on that version and I encountered the bug there. But I understand if that is not possible. Thanks for explaining @phulakun.

modasserbillah:
@phulakun @RidhaOracle can you guide me on this please? How can we get this merged? Seems like some checks are failing on unrelated stuff.

henrikingo-oracle:
The unrelated failures should have been fixed in trunk, which means if you can rebase your work onto the most recent trunk, then tests should pass.

modasserbillah:
@henrikingo-oracle I think this branch is already on top of the most recent trunk after my rebase last week. The trunk hasn't moved since then. Do you have reasons to believe otherwise?

modasserbillah:
@henrikingo-oracle if you or someone from Oracle could help me fix these failures that'd be great. These aren't related to the changes in the PR and I'm not sure how I can move this forward 🙏🏽

henrikingo-oracle:
Hi @modasserbillah 

Yes I or someone can help :-) 

I have to apologise for the mess while we are nearing normal GitHub development workflow in a couple months... But it seems there were some fixes pushed to trunk yesterday, which I think are the ones that should be of help to you as well. Can you please rebase once more and then we can see if MTR gets any greener.  Again, I really apologize for the trouble here. I thought MTR fixes were pushed already earlier.

modasserbillah:
Thanks for helping me out @henrikingo-oracle ! I have rebased it, let's see how it goes this time 🤞🏽

henrikingo-oracle:
Ok nice, it looks like the MTR tests are indeed fixed or disabled... Is it possible the format check failure is in code you edited? At least the file is part of your patch?

modasserbillah:
The format checks are part of the files I touched but not related to my changes. Do you want me to try to fix that in this PR @henrikingo-oracle?

henrikingo-oracle:
Yeah it looked like the line nr doesn't match with your patch... But I also don't see other PRs having this failure, so maybe you better look into it and see if you get this PR green :-)

ogrovlen:
> Yeah it looked like the line nr doesn't match with your patch... But I also don't see other PRs having this failure, so maybe you better look into it and see if you get this PR green :-)

My understanding is that we are here using a different version of clang-format than have been used before, so we may get errors for existing code.  The changes to this PR are later in the file than the code that is flagged, so it is very unlikely that it is caused by this PR.  That other PRs do not have this failures, is probably because they do not touch sql_table.cc

modasserbillah:
Looks like the issue was with semicolons on their own lines, I added a small formatting fix.

modasserbillah:
@henrikingo-oracle all green but merging seems to be still blocked 🤔

henrikingo-oracle:
Cool.  Nice to see some green pull requests again.
GitHub seems to be waiting for an additional reviewer. That's not me, so now we'll just wait.

modasserbillah:
@phulakun can you please request  someone with write access to review/approve this?

phulakun:
Ok, I dont have privileges to add reviewers. I will ask someone to add second reviewer for it. IIUC, you need just another review, correct?

sql/sql_table.cc
```diff
@@ -20916,9 +20914,10 @@ static bool is_any_check_constraints_evaluation_required(
...
+              !my_strcasecmp(system_charset_info, itm_fld->field_name,
                              fld.field_name) &&
-              (itm_fld->data_type() != fld.sql_type))
+              (itm_fld->data_type() != real_type_to_type(fld.sql_type)))
```
RidhaOracle, line 20920: `real_type_to_type` function only normalizes temporal types  but here we compare `Field::type()` with a real type `fld.sql_type`.

For `ENUM/SET` it becomes `MYSQL_TYPE_STRING` vs `MYSQL_TYPE_ENUM``/`MYSQL_TYPE_SET`.

Reproduction example: 
```
CREATE TABLE t (
  e ENUM('a','b'),
  CONSTRAINT ck CHECK (e IS NULL OR e IN ('a','b'))
);

ALTER TABLE t
  MODIFY e ENUM('a','b') DEFAULT 'a',
 ALGORITHM=INPLACE;
```

The ALTER only changes the default, does not affect existing row and the CHECK does not use `DEFAULT(e)` but it fails with `ER_ALTER_OPERATION_NOT_SUPPORTED`.

Should we compare `itm_fld->field->real_type()` with `fld.sql_type` instead ?

RidhaOracle: Also useful too add this test as a regression test.

PR #724, opened 2026-08-19
Add DELETE ... RETURNING clause support

evgenyp-azm:
### What does this change do?

Adds support for a RETURNING clause on single-table DELETE statements, so a DELETE can return a result set built from the rows it deleted instead of just an affected-row count. The clause goes after ORDER BY/LIMIT and accepts the same expression list as a SELECT output list, including *, qualified wildcards, aliases, and subqueries:

DELETE FROM products
  WHERE obsolete = 1
  ORDER BY created_at
  LIMIT 100
  RETURNING id, name, created_at;
  
The rows are sent to the client as an ordinary result set, protocol-identical to a SELECT.

### Why is it needed?

Returning data from modified rows is a well-established SQL pattern: PostgreSQL has supported it since 8.2, MariaDB since 10.0 for DELETE, and it appears in the SQL:2016 standard draft. It removes the need to run a SELECT before the DELETE to capture the doomed rows, which matters in three ways:

*) Audit logging and archival can capture computed or generated column values of deleted rows in one statement rather than a SELECT + DELETE round trip.
*) Queue-table and work-claiming patterns become a single atomic statement, which is the common way to hand rows to a downstream pipeline.
*) It removes a class of application-level race conditions. The SELECT-then-DELETE pattern needs explicit locking or serializable isolation to be correct under concurrency; DELETE ... RETURNING doesn't, because the read and the delete are the same operation.

### How was it tested?

- [ X ] Added/updated MTR tests under `mysql-test/t/delete_returning.test`
- [ X ] Ran the relevant full suite (name it): main,innodb

### Contributor checklist

- [ X ] I have signed the [OCA](https://oca.opensource.oracle.com) with the email on these commits
- [ X ] Commits are focused with descriptive messages

### AI assistance

- [ X ] I used AI assistance for this contribution

Anthropic Opus was used as coding assistant, in test generation and review.
Whole submitted code and tests were manually reviewed and manually tested.

### Areas touched

Parser, DELETE query executor.

### Side-effects

The change removes RETURNING_SYM from ident_keywords_unambiguous, so unquoted RETURNING can no longer be used as an identifier and must be backtick-quoted. The reasoning is in the sql_yacc.yy comment.


evgenyp-azm:
Wrong branch.

PR #733, opened 2026-08-25
Add DELETE ... RETURNING clause support

evgenyp-azm:

### What does this change do?

Adds support for a RETURNING clause on single-table DELETE statements, so a DELETE can return a result set built from the rows it deleted instead of just an affected-row count. The clause goes after ORDER BY/LIMIT and accepts the same expression list as a SELECT output list, including *, qualified wildcards, aliases, and subqueries:

DELETE FROM products

  WHERE obsolete = 1

  ORDER BY created_at

  LIMIT 100

  RETURNING id, name, created_at;

  

The rows are sent to the client as an ordinary result set, protocol-identical to a SELECT.

### Why is it needed?

Returning data from modified rows is a well-established non-standard SQL pattern: PostgreSQL has supported
it since 8.2, MariaDB since 10.0, SQLite since 3.35 (the SQL standard instead
offers data change delta tables — optional feature T495 of ISO/IEC 9075-2, added
in the 2011 edition — which nobody in the MySQL family implements). It removes the
need to run a SELECT before the DELETE to capture the doomed rows, which matters in
three ways:

*) Audit logging and archival can capture computed or generated column values of deleted rows in one statement rather than a SELECT + DELETE round trip.

*) Queue-table and work-claiming patterns become a single atomic statement, which is the common way to hand rows to a downstream pipeline.

*) It removes a class of application-level race conditions. The SELECT-then-DELETE pattern needs explicit locking or serializable isolation to be correct under concurrency; DELETE ... RETURNING doesn't, because the read and the delete are the same operation.

### How was it tested?

- [X] Added/updated MTR tests under `mysql-test/t/delete_returning.test`

- [X] Ran the relevant full suite (name it): main,innodb

### Contributor checklist

### AI assistance

Anthropic Opus was used as coding assistant, in test generation and review.

Whole submitted code and tests were manually reviewed and manually tested.

### Areas touched

Parser, DELETE query executor.

### Side-effects

The change removes RETURNING_SYM from ident_keywords_unambiguous, so unquoted RETURNING can no longer be used as an identifier and must be backtick-quoted. The reasoning is in the sql_yacc.yy comment.


evgenyp-azm:
Per request from Ridha Chahed here's the PR for DELETE .. RETURNING feature ported to trunk.
Essentially same as https://github.com/mysql/mysql-server/pull/725

roylyseng:
Thank you for the contribution!
I think it can be taken mostly as-is, however I do have a couple comments to the overall description.
You mention that DELETE ... RETURNING appears in the SQL:2016 standard draft, however I cannot see the clause in any SQL standard, including the 2026 draft standard.
OTOH, I think that RETURNING can be made a fully reserved word. The word is not available as a non-quoted identifier anyway, so there is not much of a difference.

mysql-test/t/delete_returning.test
```diff
@@ -0,0 +1,338 @@
+#
+# Tests for DELETE ... RETURNING
+#
+
+--disable_warnings
+DROP TABLE IF EXISTS t1, t2;
```
roylyseng, line 6: These statements are not necessary, as each test file starts with clean sheets.

sql/sql_delete.h
```diff
@@ -66,6 +71,9 @@ class Sql_cmd_delete final : public Sql_cmd_dml {
     optimization, use the Table_ref::updating property instead.
   */
   SQL_I_List<Table_ref> *delete_tables;
+
+  /// True if DELETE has a RETURNING clause
+  bool m_returning;
```
roylyseng, line 76: I think we can place m_returning and has_returning in class Sql_cmd_dml, with default value false.
This is a reasonable change, given that we may soon add RETURNING support for INSERT and UPDATE too.

sql/sql_yacc.yy
```diff
@@ -1366,6 +1366,17 @@ CHARSET_INFO *warn_on_deprecated_user_defined_collation(
...
+  information_schema.KEYWORDS reports it correctly (reserved=0).  However it is
+  intentionally NOT listed in ident_keywords_unambiguous - doing so creates a
+  reduce/reduce conflict with opt_delete_returning (the parser cannot tell
+  whether RETURNING after DELETE ... FROM t starts a table alias or the
+  RETURNING clause).  The practical effect is that unquoted RETURNING cannot be
+  used as an identifier; users must backtick-quote it (`returning`).  Resolving
+  the conflict properly would require significant grammar restructuring.
+  This is the smallest-impact trade-off.
+*/
 %token<lexer.keyword> RETURNING_SYM 999                 /* SQL-2016-N */
```
roylyseng: I think we can make RETURNING a fully reserved word, since there is not much difference between this definition and a full reservation. The comment above is mostly process-related and can be deleted, I guess.

sql/sql_yacc.yy
```diff
@@ -16376,7 +16397,7 @@ ident_keywords_unambiguous:
         | RESUME_SYM
         | RETAIN_SYM
         | RETURNED_SQLSTATE_SYM
-        | RETURNING_SYM
+        /* RETURNING_SYM removed, now reserved for DELETE ... RETURNING */
```
roylyseng, line 16418: Redundant comment.

sql/sp.cc
```diff
@@ -2840,6 +2841,11 @@ uint sp_get_flags_for_command(LEX *lex) {
       break;
     default:
       flags = lex->is_explain() ? sp_head::MULTI_RESULTS : 0;
+      // DELETE ... RETURNING produces a result set
+      if (lex->sql_command == SQLCOM_DELETE &&
+          lex->m_sql_cmd != nullptr &&
```
roylyseng: This check should be redundant, DELETE is always represented by an Sql_cmd object.

sql/sql_delete.cc
```diff
@@ -163,7 +164,14 @@ bool Sql_cmd_delete::precheck(THD *thd) {
   Table_ref *tables = lex->query_tables;
 
   if (!multitable) {
-    if (check_one_table_access(thd, DELETE_ACL, tables)) return true;
+    // DELETE ... RETURNING reads column data back to the client, so require
+    // SELECT privilege on the target table in addition to DELETE.
+    if (m_returning) {
+      if (check_one_table_access(thd, DELETE_ACL | SELECT_ACL, tables))
```
roylyseng, line 170: One call to check_one_table_access(), with privilege given as m_returning ? DELETE_ACL | SELECT_ACL : DELETE_ACL ?

sql/sql_delete.cc
```diff
@@ -190,6 +198,15 @@ bool Sql_cmd_delete::precheck(THD *thd) {
 bool Sql_cmd_delete::check_privileges(THD *thd) {
   DBUG_TRACE;
 
+  // DELETE ... RETURNING requires SELECT privilege on the target table.
+  if (m_returning) {
```
roylyseng, line 202: Consider to modify check_all_table_privileges() instead.
(Add SELECT_ALL if deleting table and has_returning() is true).
I also wonder if there may be a bug if delete is via a view and we have SELECT privileges for the view but not the underlying table?

sql/sql_delete.cc
```diff
@@ -415,7 +433,16 @@ bool Sql_cmd_delete::delete_from_single_table(THD *thd) {
             explain_single_table_modification(thd, thd, &plan, query_block);
         return err;
       }
-      my_ok(thd, 0);
+      if (m_returning) {
```
roylyseng, line 436: Please make a small class function to encapsulate this code.
Can we maybe send metadata up front, to avoid some of this this extra code in several places?

sql/sql_delete.cc
```diff
@@ -887,11 +957,71 @@ bool Sql_cmd_delete::prepare_inner(THD *thd) {
...
+    // RETURNING reads row data, so the user must have SELECT in addition
+    // to DELETE.  (precheck() also checks this, but re-verify here after
+    // tables are opened.)
+    Table_ref *target = select->get_table_list();
+    assert(target != nullptr);
+    // For views, target->table is nullptr (the TABLE* lives on the
+    // underlying base table); for base tables it must be set by now.
+    assert(target->is_view() || target->table != nullptr);
+    if (target == nullptr) return true;
+    if (check_single_table_access(thd, SELECT_ACL, target, false))
```
roylyseng: This check should not be necessary, ::precheck() should handle this, and ::check_privileges() will handle an execute call.

sql/sql_delete.cc
```diff
@@ -887,11 +957,71 @@ bool Sql_cmd_delete::prepare_inner(THD *thd) {
...
       return true;
   }
 
+  // Resolve RETURNING clause expressions
+  if (m_returning) {
+    // Defense-in-depth: verify SELECT privilege on the target table.
+    // RETURNING reads row data, so the user must have SELECT in addition
+    // to DELETE.  (precheck() also checks this, but re-verify here after
+    // tables are opened.)
+    Table_ref *target = select->get_table_list();
```
roylyseng, line 974: Processing of "target" is not necessary, we already have a reference in table_list.

sql/sql_delete.cc
```diff
@@ -887,11 +957,71 @@ bool Sql_cmd_delete::prepare_inner(THD *thd) {
...
+    // For views, target->table is nullptr (the TABLE* lives on the
+    // underlying base table); for base tables it must be set by now.
+    assert(target->is_view() || target->table != nullptr);
+    if (target == nullptr) return true;
+    if (check_single_table_access(thd, SELECT_ACL, target, false))
+      return true;
+
+    // Ensure name resolution context includes the delete target table,
+    // so that subqueries in RETURNING can resolve outer column references.
+    if (select->context.first_name_resolution_table == nullptr) {
```
roylyseng, line 984: I think this is always the case, so the test and assignment should not be necessary.

sql/sql_delete.cc
```diff
@@ -887,11 +957,71 @@ bool Sql_cmd_delete::prepare_inner(THD *thd) {
...
+                     /*allow_sum_func=*/false,
+                     /*split_sum_funcs=*/false,
+                     /*column_update=*/false,
+                     /*typed_items=*/nullptr,
+                     &select->fields,
+                     select->base_ref_items))
+      return true;
+
+    // Verify no aggregate functions in RETURNING
+    for (Item *item : select->fields) {
```
roylyseng, line 1006: This should not be necessary, as resolving the aggregated items will return an error.

Note also that window functions are processed the same, and this should be mentioned as a further restriction.

sql/sql_delete.cc
```diff
@@ -887,11 +957,71 @@ bool Sql_cmd_delete::prepare_inner(THD *thd) {
...
+      if (item->has_aggregation()) {
+        my_error(ER_INVALID_GROUP_FUNC_USE, MYF(0));
+        return true;
+      }
+    }
+
+    // Set up Query_result_send for sending result set to client
+    Prepared_stmt_arena_holder ps_holder(thd);
```
roylyseng, line 1014: A Prepared_stmt_arena_holder is not necessary, this code always run in the resolver which is against (prepared) statement mem_root.

mysql-test/suite/json/t/json_value.test
```diff
@@ -277,9 +277,12 @@ SELECT JSON_VALUE(json_value, '$.a') AS json_value FROM json_value;
 DROP TABLE json_value;
 
 # RETURNING is a non-reserved word both in the standard and in MySQL.
```
roylyseng, line 279: Replace with "RETURNING is non-reserved word in SQL standard but reserved in MySQL" ?

mysql-test/t/delete_returning.test
```diff
@@ -0,0 +1,338 @@
...
+
+DELETE FROM t3 PARTITION (p0) WHERE a=4 RETURNING *;
+SELECT * FROM t3 WHERE a=4;
+DROP TABLE t3;
+
+--echo #
+--echo # Test 25: EXPLAIN DELETE ... RETURNING (no data returned, no rows deleted)
+--echo #
+--replace_column 10 X
+EXPLAIN DELETE FROM t1 WHERE a=2 RETURNING *;
```
roylyseng, line 242: If you add "format=tree" to EXPLAIN, this test file will also run well with --hypergraph.

evgenyp-azm:
You're right about the standard, haven't checked it properly before posting.
Re reserved:Pg, MariaDB and SQLite has RETURNING as a reserved word. So while it makes sense to be consistent,  it's a breaking change.
I'll update the PR with making it reserved.

mysql-test/r/delete_returning.result
```diff
@@ -0,0 +1,515 @@
...
+3	ccc
+4	dddd
+1	A
+4	DDDD
+5	EEEEE
+7	GGGGGGG
+2	BB
+2	bb
+#
+# Test 6: DELETE ... RETURNING with aggregate function (error)
```
roylyseng, line 94: I think we need a similar test with a window function.

sql/sql_delete.cc
```diff
@@ -190,6 +198,15 @@ bool Sql_cmd_delete::precheck(THD *thd) {
 bool Sql_cmd_delete::check_privileges(THD *thd) {
   DBUG_TRACE;
 
+  // DELETE ... RETURNING requires SELECT privilege on the target table.
+  if (m_returning) {
+    Table_ref *const table_list = lex->query_block->get_table_list();
+    assert(table_list != nullptr);
+    if (table_list == nullptr) return true;  // Fail-closed if unexpectedly null
```
roylyseng, line 205: I am not sure this will help anything, because the contract for the function is that with a true return, there should be a diagnostics value. I think it is better to delete this statement.

sql/sql_delete.cc
```diff
@@ -887,11 +957,71 @@ bool Sql_cmd_delete::prepare_inner(THD *thd) {
 
     assert(!select->group_list.elements);
     if (select->setup_base_ref_items(thd)) return true; /* purecov: inspected */
-    if (setup_order(thd, select->base_ref_items, &tables, &select->fields,
+    // When RETURNING is present, select->fields contains unresolved RETURNING
+    // items. ORDER BY in DELETE resolves against table columns, not RETURNING,
+    // so pass an empty field list.
+    mem_root_deque<Item *> empty_fields(thd->mem_root);
+    if (setup_order(thd, select->base_ref_items, &tables,
```
roylyseng, line 962: I think it will be reasonable to process RETURNING fields before ORDER BY. That way, all fields in "fields" will be resolved, and we don't need the empty list.
Resolving ORDER BY will also populate the empty list with items, which are subquently abandoned. Probably not a good idea...
OTOH, resolving RETURNING before ORDER BY will make us add aliases in the select list, which can later be picked up by ORDER BY, as in:

  DELETE FROM t1 WHERE a=1 ORDER BY c RETURNING a+b AS c;

I guess this is an acceptable change.

roylyseng (changes requested):

sql/sql_delete.cc
evgenyp-azm: I'd prefer to keep metadata sending place as is, to be consistent with overall statement handling. Also, there are only couple places which sent metadata, so benefits are questionable IMHO.

evgenyp-azm:
Addressed review comments.

PR #735, opened 2026-08-27
Bug#121045 Server telemetry metrics backed by aggregated_stats_buffer always report 0

hunter-kang:
### What does this change do?

Fixes [Bug#121045](https://bugs.mysql.com/bug.php?id=121045). 25 server telemetry (OTEL) metrics under `mysql.stats` / `mysql.stats.handler` always export 0: their per-row producers increment only the per-THD `status_var`, never the sharded `aggregated_stats_buffer` that the metric callback (`get_metric_aggregated_integer`) reads. This adds the missing shard write next to each per-THD store, so the metrics report real activity.

### Why is it needed?

Without it, 25 metrics (the `Handler_*` read/write family, plus `Select_*`, `Sort_*`, `Created_tmp_*`, `long_query_count`, `max_execution_time_exceeded`) read a constant 0 under real load. Introduced in 8.2.0 by WL#15199 ([commit 83926c7](https://github.com/mysql/mysql-server/commit/83926c7fda58664b649f0731a973ad610985d36e)), which added the sharded buffer and its OTEL reader but wired the shard write only for transaction-level counters (e.g. `ha_commit`), not the per-row producers. It went unnoticed because `SHOW STATUS` sums the per-THD values through a different path, so the counters look correct there.

### How was it tested?

- [x] Ran the relevant full suite (name it): **perfschema**

New test `perfschema.telemetry_metrics_shard_bug` drives the Select / Sort / Created_tmp / Handler families and asserts each covered metric is populated. It fails on the unfixed server (metrics read 0) and passes with the fix. The full `perfschema` suite was run (618/624 pass, including this test); the handful of failures are pre-existing or parallel-execution-flaky (verified: they fail on the unpatched baseline or pass when run in isolation), none caused by this change.

This change adds one `memory_order_relaxed` atomic per handler call on the per-row path. sysbench `oltp_read_only` and `oltp_write_only` (baseline vs. patched) showed no measurable throughput or latency regression.

### Contributor checklist

### AI assistance

Used Claude Opus 4.8 (via Claude Code) throughout this contribution: it generated the code change and the MTR test (iterated across revisions), ran the sysbench performance testing, researched the root cause and the commit that introduced the bug (WL#15199, 8.2.0), and drafted this PR description. All of it was human-reviewed and verified: I manually checked the code and its call paths, and manually confirmed the MTR test's before/after behavior (local builds + MTR).

### Areas touched

`sql` (handler, sql_class, log) and the storage-engine handlers (innodb, archive, csv, federated, heap, myisam, myisammrg, ndb, perfschema, temptable); `perfschema` MTR suite.

### Note on later versions

This fix targets 8.4 and covers the 25 affected counters present there. One further counter, `count_hit_tmp_table_size`, exists only in 9.x (added after 8.4) and has the same defect; it is not touched by this change and will need to be handled separately.

### Copyright


marcalff (commented):
Thanks for the fix.

The analysis in the bug report, and the fix in this PR, looks correct,
updating the global shards was indeed missing for some metrics.

The next steps for this PR are:

- deploy CI in github for branch 8.4 and 9.7 (only 26.7 has CI),
- and/or try the patch internally for now using the old infrastructure.

and then:

- merge to 8.4-LTS, 9.7-LTS and trunk, exact details not known at this point.

All this is pending completion of the migration to github.

Related, your github account shows as first contributor to this repository, so we should make sure when applying and merging the patch, that you get proper credits, and that github detects your account as a contributor from now on.

Again, pending completion of the migration to github.

PR #736, opened 2026-08-28
Bug#121063 Members of one group assign different GNOs to the same transaction

matias-sanchez:
Bug#121063: https://bugs.mysql.com/bug.php?id=121063

## What happens

Two members of the same group assign different GNOs to the same transaction, in multi-primary mode, when a member leaves and rejoins during a rolling restart under write load. The group splits into internally consistent sets that disagree on the GTID of the same transaction. It can surface on the `group_replication_applier` channel as Error_code 1032 or 1062, or stay silent with every member ONLINE and the disagreement present only in the binary logs.

Reproduced on 8.4.8, 8.4.11 and 9.7.2. A self contained reproducer is attached to the bug.

## Where it starts

A reserved synode is only ours while we still hold the node index it was reserved under.

`local_synode_allocator` stamps the member's current index into the synode, `synode.node = my_nodeno`. Still inside `reserve_synode_number`, the task yields in the `while (too_far(*msgno))` loop at `TIMED_TASK_WAIT`. During that yield, `site_install_action` reassigns `site->nodeno`. The reservation still carries the old index, so `proposer_task` brands and proposes into a slot that now belongs to another node.

The header comment of `xcom_base.cc` states the rule at line 107: only node N may propose a value for synode {X N}. With two proposers on the same slot at `cnt=0`, `acceptor.promise` is never raised, since it is assigned in exactly one place, inside `handle_simple_prepare`, which is the phase 1 decision. Both proposals are accepted, both are learned, and `handle_learn` keeps whichever LEARN arrived first because of the `/* Avoid re-learn */` guard. Members that heard different values first deliver different payloads.

A bpftrace trace of one such slot followed from ACCEPT through LEARN to DELIVER is in the bug, posted 28 Aug 2026.

## The change

Before proposing, verify the reservation still carries the member's own node index. If it does not, drop it through `retry_new` and take a new one.

This is the same check `incr_msgno` (`xcom_base.cc:668`) already makes whenever it advances, with the comment `In case site and node number has changed`. The client transaction is not lost, it goes out in a slot that does belong to the member.

## Testing

In the lab the anomaly goes to zero: 0 stale proposals out of 15,523,048, against 993 predicted from the unpatched rate, and 0 divergences out of 5,667,504 transactions, against 17.1 expected. No run carrying the patch has reproduced the divergence, on two hosts.

Performance: 12 runs per arm across two hosts, load only, no restarts, no emulated latency, comparing the patched plugin against the same source built without the patch and against the stock image. Ratio of median successful operations per 10 s window, bootstrap over runs: patched over unpatched [0.9998, 1.0004], patched over stock [0.9994, 1.0003]. A positive control with a known throttle was detected at [0.9944, 0.9953], so the method resolves differences of about 0.5%. The workload is rate limited, so this measures behaviour under a production-like load and not peak capacity.

I do not have an MTR test for this. The reproduction needs seven members, rolling restarts and sustained write load, which does not fit the standard framework. The reproducer attached to the bug builds that environment and decides the verdict by decoding every member's binary log and comparing all 21 pairs.

### DIFF
--- a/plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
+++ b/plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
@@ -1756,6 +1756,15 @@ static void push_msg_3p(site_def const *site, pax_machine *p, pax_msg *msg,
              STRLIT(pax_op_to_str(msg->op)));
 }
 
+/* A reserved synode is only ours while we still hold the node index it was
+   reserved under. A view change that renumbers us hands that slot to another
+   node, which may reserve it as well. */
+static bool_t reservation_is_stale(synode_no msgno) {
+  site_def const *site = find_site_def(msgno);
+  node_no me = site ? get_nodeno(site) : VOID_NODE_NO;
+  return me != VOID_NODE_NO && me != msgno.node;
+}
+
 /* Brand client message with unique ID */
 static void brand_client_msg(pax_msg *msg, synode_no msgno) {
   assert(!synode_eq(msgno, null_synode));
@@ -2519,6 +2528,13 @@ static int proposer_task(task_arg arg) {
     }
 
     brand_client_msg(ep->client_msg->p, ep->msgno);
+
+    /* Only a locally allocated synode carries our own node index; a remote or
+       global allocation carries the allocating leader's, by design. */
+    if (ep->synode_allocation == synode_allocation_type::local &&
+        reservation_is_stale(ep->msgno)) {
+      GOTO(retry_new);
+    }
 
     for (;;) { /* Loop until the client message has been learned */
       /* Get a Paxos instance to send the client message */
--- a/plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
+++ b/plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
@@ -2529,13 +2529,6 @@ static int proposer_task(task_arg arg) {
 
     brand_client_msg(ep->client_msg->p, ep->msgno);
 
-    /* Only a locally allocated synode carries our own node index; a remote or
-       global allocation carries the allocating leader's, by design. */
-    if (ep->synode_allocation == synode_allocation_type::local &&
-        reservation_is_stale(ep->msgno)) {
-      GOTO(retry_new);
-    }
-
     for (;;) { /* Loop until the client message has been learned */
       /* Get a Paxos instance to send the client message */
 
@@ -2546,8 +2539,8 @@ static int proposer_task(task_arg arg) {
         goto retry_new;
       }
 
-      /* Checked again after wait_for_cache: that call can suspend, and a view
-         change during the wait leaves the reservation stale. */
+      /* wait_for_cache can suspend, and a view change during the wait leaves
+         the reservation stale. */
       if (ep->synode_allocation == synode_allocation_type::local &&
           reservation_is_stale(ep->msgno)) {
         GOTO(retry_new);
--- a/plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
+++ b/plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
@@ -1759,7 +1759,7 @@ static void push_msg_3p(site_def const *site, pax_machine *p, pax_msg *msg,
 /* A reserved synode is only ours while we still hold the node index it was
    reserved under. A view change that renumbers us hands that slot to another
    node, which may reserve it as well. */
-static bool_t reservation_is_stale(synode_no msgno) {
+bool_t reservation_is_stale(synode_no msgno) {
   site_def const *site = find_site_def(msgno);
   node_no me = site ? get_nodeno(site) : VOID_NODE_NO;
   return me != VOID_NODE_NO && me != msgno.node;
@@ -2544,6 +2544,13 @@ static int proposer_task(task_arg arg) {
         G_INFO("Could not get a pax_machine for msgno %lu. Retrying",
                (unsigned long)ep->msgno.msgno);
         goto retry_new;
+      }
+
+      /* Checked again after wait_for_cache: that call can suspend, and a view
+         change during the wait leaves the reservation stale. */
+      if (ep->synode_allocation == synode_allocation_type::local &&
+          reservation_is_stale(ep->msgno)) {
+        GOTO(retry_new);
       }
 
       assert(ep->p);
--- a/unittest/gunit/libmysqlgcs/xcom/gcs_xcom_stale_reservation-t.cc
+++ b/unittest/gunit/libmysqlgcs/xcom/gcs_xcom_stale_reservation-t.cc
@@ -35,46 +35,39 @@ namespace xcom_stale_reservation_unittest {
 class XcomStaleReservation : public GcsBaseTest {
  protected:
   void SetUp() override {
-    m_addr = new std::string("127.0.0.1:12345");
-    char const *names[]{m_addr->c_str()};
-    m_na = new_node_address(1, names);
-
     /* The view in force when the synode is reserved: this member is node 1. */
-    m_before = new_site_def();
-    init_site_def(1, m_na, m_before);
-    m_before->start = m_synode_before;
-    m_before->nodeno = 1;
-    push_site_def(m_before);
+    char const *names[]{"127.0.0.1:12341", "127.0.0.1:12342",
+                        "127.0.0.1:12343"};
+    node_address *na = new_node_address(3, names);
 
-    /* The view installed while the reservation is held: the member is now
-       node 0. */
-    m_after = new_site_def();
-    init_site_def(1, m_na, m_after);
-    m_after->start = m_synode_after;
-    m_after->nodeno = 0;
-    push_site_def(m_after);
+    site_def *before = new_site_def();
+    init_site_def(3, na, before);
+    before->start = synode_no{1, 10, 0};
+    before->nodeno = 1;
+    push_site_def(before);
+    delete_node_address(3, na);
   }
 
-  void TearDown() override {
-    push_site_def(nullptr);
-    free_site_defs();
-    delete_node_address(1, m_na);
-    delete m_addr;
+  /* The view installed while the reservation is held: the first member is
+     gone, so this one is now node 0. */
+  void install_new_view() {
+    char const *names[]{"127.0.0.1:12342", "127.0.0.1:12343"};
+    node_address *na = new_node_address(2, names);
+
+    site_def *after = new_site_def();
+    init_site_def(2, na, after);
+    after->start = synode_no{1, 20, 0};
+    after->nodeno = 0;
+    push_site_def(after);
+    delete_node_address(2, na);
   }
 
-  std::string *m_addr{nullptr};
-  node_address *m_na{nullptr};
-  site_def *m_before{nullptr};
-  site_def *m_after{nullptr};
-
-  /* A view is active from its start synode on, so a slot below 20 falls under
-     the old view and one at or above it under the new one. */
-  synode_no const m_synode_before{1, 10, 0};
-  synode_no const m_synode_after{1, 20, 0};
+  void TearDown() override { free_site_defs(); }
 };
 
 /* Under the view it was taken in, the reservation is ours. */
 TEST_F(XcomStaleReservation, reservation_under_the_current_index_is_fresh) {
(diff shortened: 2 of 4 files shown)


plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
```diff
@@ -2520,6 +2529,13 @@ static int proposer_task(task_arg arg) {
 
     brand_client_msg(ep->client_msg->p, ep->msgno);
 
+    /* Only a locally allocated synode carries our own node index; a remote or
+       global allocation carries the allocating leader's, by design. */
+    if (ep->synode_allocation == synode_allocation_type::local &&
```
jujose-1, line 2534: This check covers a view change while `reserve_synode_number()` is suspended, which matches the reported trace. 

However, `wait_for_cache()` below is another `TASK_CALL` and may suspend at `TIMED_TASK_WAIT`. If the member is renumbered during that wait, the reservation can become stale after this check. 

Could we revalidate the local reservation immediately after `wait_for_cache()` succeeds—after the null check and before locking or modifying the Paxos machine? 

Keeping both checks would avoid waiting on an already stale reservation while also establishing ownership after the final possible yield.

plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
jujose-1, line 1: Could you please add a focused XCom unit test that simulates a locally reserved synode followed by a view change that changes the member’s node index, and verifies that the reservation is detected as stale?

jujose-1 (commented):
Hi Matias, 
Thank you for the contribution and for working on this fix. I’ve left a few review comments for your consideration. Please take a look when you have a chance.

Regards,
Justin Jose

plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
matias-sanchez: Hi @jujose-1 huge thanks on your review.
Good catch, thanks. Added the revalidation right after wait_for_cache(), after the null check, and kept the first check as you suggested.
I tested this over a 3 hours runs with rolling restarts under load and I didn't see any meaningful impact on throughput, with the GNO mismatch not reproducing. Feels safe.

plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
matias-sanchez: Added in unittest/gunit/libmysqlgcs/xcom/gcs_xcom_stale_reservation-t.cc. It sets up a view where a member is node 1, then a later view where it is node 0, and asserts the reservation taken under the old index is reported stale. Three more cases cover the fresh ones.

matias-sanchez:
Hi @jujose-1, thanks for taking the time to review this and for the comments. Added the requested changes.

unittest/gunit/libmysqlgcs/xcom/gcs_xcom_stale_reservation-t.cc
jujose-1: Thanks for adding the test. Could you please extend it to cover the retry path in `proposer_task()` as well? Currently, the test still passes if the two checks that route a stale reservation through `retry_new` are removed.

A deterministic test hook could exercise a sequence conceptually similar to this:
```

  install_view(/* local node index = */ old_index);
  enqueue_client_message(client_message);

  suspend_proposer_after_local_reservation();
  synode_no const stale_synode = proposer_reserved_synode();

  install_view(/* local node index = */ new_index);
  resume_proposer();

  EXPECT_FALSE(was_proposed(stale_synode));
  EXPECT_EQ(new_index, retried_proposal_synode().node);
  EXPECT_EQ(client_message, retried_client_message());

```
The key verification is that the stale synode is not proposed and that the same client transaction is retried using the current local node index.

The helper names above are illustrative; the main goal is to cover the stale-reservation-to-retry_new path explicitly.

unittest/gunit/libmysqlgcs/xcom/gcs_xcom_stale_reservation-t.cc
```diff
@@ -0,0 +1,100 @@
...
+  void SetUp() override {
+    m_addr = new std::string("127.0.0.1:12345");
+    char const *names[]{m_addr->c_str()};
+    m_na = new_node_address(1, names);
+
+    /* The view in force when the synode is reserved: this member is node 1. */
+    m_before = new_site_def();
+    init_site_def(1, m_na, m_before);
+    m_before->start = m_synode_before;
+    m_before->nodeno = 1;
```
jujose-1: The old site is initialized with one member by `init_site_def(1, ...)`, so node index 1 appears to be outside its valid range. Also, both site definitions are installed during `SetUp()`, so the test does not seem to observe the same reservation across the view installation.

Could we use valid memberships instead—for example, an old view `[removed, local, other]`, where the local member has index 1, followed by a new view `[local, other]`, where it has index 0?

The renumbering test could then verify that the same reserved synode is fresh before installing the new view and stale afterward.

jujose-1 (commented):
Hi @matias-sanchez ,

Thank you for updating the patch and addressing the earlier feedback. The production change looks good to me. I’ve left one additional comment regarding the unit-test setup.

I’ll also wait for @tiagoportelajorge 's review and any additional feedback he may have.

Thank you again for working on this.

Regards,
Justin Jose

kamil-holubicki:
My proposal is to add two debug asserts:

1. Proposer side (offending node) in `prepare_push_2p()` after the ballot setup: `assert(p->proposer.bal.node == p->synode.node)`. The node can propose only for its synode. Detect the case when the current node index mismatches the index at the synode reservation time.

2. Acceptor side in `handle_simple_accept()`: `assert(m->proposal.cnt != 0 || m->proposal.node == m->synode.node);`. In fast path (cnt==0) the node can propose only for its synode.

unittest/gunit/libmysqlgcs/xcom/gcs_xcom_stale_reservation-t.cc
matias-sanchez: Hi @jujose-1, thanks for taking the time to explain how to set this up properly!. I just pushed a new commit that reworks the test: the old view is now `[removed, local, other] `with the local member at index 1, the new view is `[local, other] `where it has index 0, and SetUp() installs only the old view. The renumbering test now checks that the same reserved synode is fresh before the new view is installed and stale afterward. Last commit touches the test only.

matias-sanchez:
Thanks @kamil-holubicki . However, I'm not sure on how to safely add these asserts, because the issue covered is on multi primary and I'm not sure how to add that assert so that it does not fire when `paxos_single_leader` is ON, as in that mode the synode carries the leader's node index, not the proposer's. @jujose-1  would adding `site->max_active_leaders != active_leaders_all` be the proper way to validate this? Also, do you recommend adding this on this PR or would you prefer a follow up one?

jujose-1:
Thanks, @matias-sanchez , for updating the unit test. The revised test now uses valid memberships and verifies the same reservation across the view change. This looks good to me.

Thanks also, @kamil-holubicki, for the suggestions.

> @jujose-1 would adding `site->max_active_leaders != active_leaders_all` be the proper way to validate this? Also, do you recommend adding this on this PR or would you prefer a follow up one?

My understanding is that the proposed assertions would cover the scenario reported in this bug, but they may not hold for remotely allocated synodes, where the synode can legitimately carry the allocating leader’s node index rather than the proposer’s.

Given that the reported issue involves a locally allocated reservation, and the current fix scopes the ownership validation to that path, I suggest keeping this PR focused on that change. The acceptor side does not currently have the allocation provenance needed to apply the same invariant safely, so I would prefer not to add the proposed assertions here.

plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
tiagoportelajorge: Hi!

Just to add a note here... the loop that follows is the main loop of the proposer, where each of the 8 proposers are waiting for the message to be accepted.

In short, I don't think that this check makes sense here, and maybe the one inside the loop is enough, because even with view changes, we keep the proposals active. They will end up being accepted in another configuration, but it will avoid the said problem.

plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
```diff
@@ -1756,6 +1756,15 @@ static void push_msg_3p(site_def const *site, pax_machine *p, pax_msg *msg,
              STRLIT(pax_op_to_str(msg->op)));
 }
 
+/* A reserved synode is only ours while we still hold the node index it was
+   reserved under. A view change that renumbers us hands that slot to another
+   node, which may reserve it as well. */
+bool_t reservation_is_stale(synode_no msgno) {
+  site_def const *site = find_site_def(msgno);
```
tiagoportelajorge, line 1763: I need to think a bit more about this, because I don't know exactly how `find_site_def` behaves for synodes that were still in the proposal phase. It works well for synodes that are already in the stream and accepted by the group.

That synode might still happen, but now necessarily with the data you are proposing. That is why we have checks like `if (match_my_msg(ep->p->learner.msg, ep->client_msg->p))`

tiagoportelajorge:
HI @matias-sanchez , thank you for working on this!

I've reopened a conversation because I think that the code can be simplified. I've also dropped some food for thought, even for me to double-check the behaviour of `find_site_def`

matias-sanchez:
Thanks @jujose-1  and @tiagoportelajorge for the review. @tiagoportelajorge  I will look deeper into the code on the points you raised and update soon, as at first glance I am still not sure how it could be simplified, so let me know any further thoughts or guidance you may have.

tiagoportelajorge:
> Thanks @jujose-1 and @tiagoportelajorge for the review. @tiagoportelajorge I will look deeper into the code on the points you raised and update soon, as at first glance I am still not sure how it could be simplified, so let me know any further thoughts or guidance you may have.

Hi @matias-sanchez,

Please check my comment in the reopened thread:

> Hi!
> Just to add a note here... the loop that follows is the main loop of the proposer, where each of the 8 proposers are waiting for the message to be accepted.
> In short, I don't think that this check makes sense here, and maybe the one inside the loop is enough, because even with view changes, we keep the proposals active. They will end up being accepted in another configuration, but it will avoid the said problem.

Regards,

Tiago

plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
matias-sanchez: Hi @tiagoportelajorge, thanks, you are right!. I tested it: I built the plugin with this check removed and ran the same workload with rolling restarts, and the divergence did not come back. You are right, the one inside the loop is enough. I just pushed a commit removing this one.

matias-sanchez: thanks for pointing at this

matias-sanchez:
Hi @tiagoportelajorge , thanks for reposting your comment as I had missed that one. I tested you're proposal and you were right, that check was not needed, the one inside the loop is enough :) . Just pushed a commit removing it and leaving only the one inside the loop.

plugin/group_replication/libmysqlgcs/src/bindings/xcom/xcom/xcom_base.cc
matias-sanchez: Hi Tiago. As far as I can tell `find_site_def` does not treat the two cases differently: it walks the `site_defs` list comparing the synode against each site's start, and has no notion of whether the synode was proposed, accepted or executed.

On `match_my_msg`, I think it catches a different scenario: it runs after finished(p), so the value is already learned and kept by handle_learn's re-learn guard by then.

Let me know if you think I might be missing something. It is tough source code i'm not entirely familiar with.

matias-sanchez:
Hi @tiagoportelajorge @jujose-1, only posting back to check with you if something is still pending at my side on this PR, or are we good?

jujose-1:
Hi @matias-sanchez, Thank you for checking in. There is nothing pending from your side at the moment. We are having an internal discussion on the process for integrating the change into trunk.

We’ll let you know if anything further is needed from your side. Thank you again for the contribution.

Regards,
Justin Jose

matias-sanchez:
Hi @jujose-1 thanks u so much for the reply! cheers

Issue #684, opened 2026-06-26
In-Memory Relaylog


HugoWenTD:
This proposal is authored by the AWS RDS Aurora MySQL team. We were encouraged by the Change Stream Applier (CSA) work Oracle presented on April 29, and in this discussion we propose optimizations to further improve CSA's replication performance. We are looking to collaborate with the community and implement these improvements in upstream MySQL. The tech contact is Xueting Wu (@ams-binlog-replication).

## Introduction

Oracle recently introduced a MySQL Labs Release featuring the Change Stream Applier (CSA), a new execution model for the MySQL logical replication applier that significantly improves SQL thread parallelism and throughput. At AWS Aurora MySQL, we have been independently implementing equivalent optimizations in Binlog, including parallelizing the deserialization of binlog events, improving scheduling by enabling worker threads to make partial progress during transaction execution, and parallelizing the application of large transactions, work that overlaps considerably with CSA's current and future scope.

We have conducted experiments with CSA and observed throughput improvements in the SQL subsystem (coordinator + worker threads) of up to 2×. We believe additional optimizations can push IO thread performance to 2x, and SQL thread by 1.5×. This document presents those ideas for your consideration.

## CSA Performance and Bottlenecks

We benchmarked CSA against the baseline MySQL 9.7's multi-threaded applier replication under identical workload conditions. All configurations used 64 worker threads, GTID enabled, binary logging disabled on the replica, and flush-at-commit (`flush_trx = 1`). Commit order and performance schema were disabled on the replica to reduce overhead.

**Throughput metric:** We generate a sustained Sysbench workload on the source and measure the rate of binlog bytes processed per second (MB/s) at each stage of the replication pipeline. To measure IO throughput, we pause the SQL thread and let the IO thread fill the relay log unimpeded. To measure SQL throughput, we pause the IO thread after the relay log has been populated and measure how long the SQL subsystem takes to process the buffered events. E2E throughput is the effective replication rate from source to committed replica state with both threads running concurrently.

| Configuration | IO Throughput | SQL Throughput | E2E Throughput |
|---|---|---|---|
| MySQL 9.7 Source | N/A | N/A | 56 MB/s |
| MySQL 9.7 Replica – no CSA | 40 MB/s | 23 MB/s - 24.4 MB/s | 22.2 MB/s |
| MySQL 9.7 Replica - CSA | 40 MB/s | 40 – 47.3 MB/s | 36.6 - 39.7 MB/s |

[...]

PR #743, opened 2026-09-03
Feature #684: In-memory Relaylog

yangxt-aws:
- Primary issue: #684
- Design: `design/replication/684-in-memory-relaylog`

### How was it tested?

- [x] Ran the relevant full suite (name it): `rpl`, `rpl_nogtid`, `rpl_gtid`.

### Contributor checklist

- [ ] Commits are focused with descriptive messages

### AI assistance

AI is used for
* code generation
* test generation
* refactoring
* review

### Areas touched
replication


share/messages_to_clients.txt
```diff
@@ -11055,17 +11052,14 @@ ER_CSA_CRST_REQUIREMENT_GTID_ONLY
 ER_DA_CANNOT_REPLICATE_WITHOUT_BINLOG
   eng "Cannot replicate from source as it does not have logical log enabled. Check source configuration to enable it."
 
-ER_REPLICA_IDENTITY_CONFLICT
-  eng "A dump thread for replica UUID '%s' or server ID %u already exists under a different authenticated account."
-
-ER_BULK_READER_PATH_IS_DIRECTORY
-  eng "Bulk Reader: Path is a directory, expected a file: %s"
+ER_CRST_IN_MEMORY_RELAYLOG_ONLY_FOR_CSA
```
wxueting-aws, line 11070: There is merge conflicts on this file, please resolve

yangxt-aws: Thank you, already resolved.

yangxt-aws:
To whom it may concern, the published code change in this PR is for demonstration and testing only.
It includes:
* CRST for IMR.
* IMR normal memory path.
* IMR large transaction spill path.
* IMR IO/SQL recovery.
* Unittests + MTR tests

The code is for validation and map the design to the code. More changes may be submitted in the following development process.

sql/changestreams/apply/storage/relay_log/cached_event_payload.cpp
```diff
@@ -37,6 +37,16 @@ Cached_event_payload::Cached_event_payload(const Event_payload &payload,
   assert(m_fde_ptr != nullptr);
 }
 
+Cached_event_payload::~Cached_event_payload() {
```
karolina-szczepankiewicz, line 40: Great. Would you like to report this as a bug and submit the fix as a separate contribution / a separate PR?

design/replication/684-in-memory-relaylog/benchmark-results.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Results
+
+Throughput in MB/s (higher is better).
+
+| Stage      | CSA applier (MB/s) | CSA + In-memory relaylog (MB/s) |
+| ---------- | -----------------: | ------------------------------: |
+| Receiver   |                 80 |                             230 |
+| Applier    |                 66 |                              67 |
+| End-to-end |                 44 |                              65 |
```
karolina-szczepankiewicz, line 24: Great improvement!

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+
+Applier (SQL thread)
+
+The coordinator consumes transactions through the existing `Transaction_provider` interface, so the `Csa_service::run` loop, scheduler, dependency tracking, and worker pool are all reused unchanged. The changes on the coordinator side are limited to the `Reader` interface.
+A new reader `Queued_transaction_reader` is introduced as a memory path counterpart of `Relay_log_adaptive_reader`. Its `read()` takes the envelope at `dispatch_seqno` from the `Trx_envelope_queue`, advances the cursor, and wraps the envelope's `Fetchable_transaction` into a fresh `Job_applier` using the same constructor the relaylog reader uses today.
+Everything downstream is untouched. The coordinator obtains the next transaction from the provider, computes scheduling dependencies from the logical clock and commit-order inputs, and dispatches it to the worker thread pool exactly as today.
+The coordinator also tries to advance `commit_seqno` in the queue over a contiguous run of committed transactions at the head and dequeues each. A transaction that commits behind an uncommitted head is only marked committed and swept later when the head commits.
```
karolina-szczepankiewicz, line 67: I had a comment about the previous approach, but you changed it to lazy advancement of the commit LWM by the coordinator, which is great. The only remaining item is a short explanation of why the queue maintains a committed prefix separately from the CSA scheduler's existing `Clock_lwm_registry`, which also computes a contiguous completed task prefix regardless of the value of `replica_preserve_commit_order`.

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
+### Functional Requirements
+
+- FR1. The feature must provide a per-channel option, introduced as a new `CHANGE REPLICATION SOURCE TO` option `IN_MEMORY_RELAYLOG_ENABLED`, that controls whether the feature is enabled. It is `ON` by default for eligible channels. When `OFF`, replication must fall back to the standard disk path relaylog.
```
karolina-szczepankiewicz, line 3: The implementation defaults `IN_MEMORY_RELAYLOG_ENABLED` to `OFF`. Please update the design to use `OFF` consistently. With this default, IMR is enabled only through an explicit `CHANGE REPLICATION SOURCE TO` statement, and no separate default is needed for upgraded channels.

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+- FR37.1. The error-diagnostic file must be created under `#in_memory_relaylog_temp_files` in relaylog format. There is only one error diagnostic file per channel, and it must survive startup cleanup. It's removed only when the replication channel restarts or the channel is reset.
+- FR38. An error diagnostic file created by the feature must have a name matching the pattern `imr_err_<channel>_<unique_id>`, where `<channel>` is a lowercase channel identifier and `<unique_id>` is a lowercase identifier unique within `#in_memory_relaylog_temp_files` (imr_err stands for "In-Memory Relaylog Error").
+- FR39. Startup cleanup must not delete temp error diagnostic files, contrary to FR25.
+- FR40. If the feature encounters a non-recoverable error, the user must be able to fall back to the standard disk path relaylog and resume replication without data loss by stopping the channel, setting `CHANGE REPLICATION SOURCE TO` with `IN_MEMORY_RELAYLOG_ENABLED = OFF`, and restarting it.
+
```
karolina-szczepankiewicz, line 65: It's good that FR30-FR30.2 now cover the partial transaction on STOP REPLICA IO_THREAD. The remaining thing is disabling IMR through CRST. Please add one functional requirement stating that the threads must be stopped, the IMR queue is removed, and the next start uses the disk relay log. Please also state what happens to spill and diagnostic files.

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+trx_length > disk_path_threshold      -> spill path
+queue_bytes + trx_length <= limit     -> memory path
+otherwise                             -> block until commits free space, then memory path
+```
+
+Reserving the exact `trx_length` up front guarantees a memory-path transaction fits before streaming begins. A transaction larger than the `IN_MEMORY_RELAYLOG_SPILL_THRESHOLD` always takes the spill path.
+
+Thread safety
+The queue uses a three-tier locking structure so that its shared structure is safe while per-transaction streaming stays parallel.
+A single queue-level mutex guards the deque, the three cursors, and envelope properties. Every structural mutation requires this mutex, so the receiver, coordinator, and workers never modify the queue concurrently. These critical sections are small and are taken only at transaction boundaries.
```
karolina-szczepankiewicz, line 57: After the latest changes, my understanding is that workers do not access the queue structure or its cursors. On commit, a worker only updates the dispatched envelope's completion state and releases its payload. The coordinator later advances commit_seqno and removes entries. Please clarify this here. The current text suggests that workers acquire the queue mutex

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+A per-envelope mutex guards each envelope's status flags (committed / truncated) and its payload reference.
+The per-transaction event streaming is synchronized separately, by `Fetchable_transaction`'s own mutex and condition variable. This mechanism already exists in the current code, so different transactions stream fully in parallel.
+The memory usage counter is an atomic variable. Modifying the variable is a lock-free update. A mutex plus a condition variable is used only when the memory counter is at the `IN_MEMORY_RELAYLOG_LIMIT` and the receiver must block until commits free enough memory.
```
karolina-szczepankiewicz, line 60: Please consider using `Resource_instance_monitor` for `IN_MEMORY_RELAYLOG_LIMIT`. It already provides lightweight reservation, waiting, cancellation, and release. The optimistic reservation path uses atomic operations without taking a mutex or entering the wait loop. Ofc, this should be a separate resource from `APPLIER_EVENT_MEMORY_LIMIT`.

design/replication/684-in-memory-relaylog/high-level-description.md
```diff
@@ -0,0 +1,60 @@
...
+
+`Group Replication` channel has no IO thread. So the streaming and recovery logic is not feasible in GR setup.
+
+With In-memory relaylog, file based relaylog-related surfaces are reduced.
+
+`SHOW REPLICA STATUS` relaylog position fields (`Relay_Log_File`, `Relay_Log_Pos`, `Relay_Log_Space`) are not applicable
+
+`sync_relay_log` is not applicable
+
+`SHOW RELAYLOG EVENTS` is not applicable
```
karolina-szczepankiewicz: Please state the exact result of this command

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+- FR16.1. When a CRST clause sets `IN_MEMORY_RELAYLOG_LIMIT` alone, the server must validate that the provided value: (1) is greater than the current `IN_MEMORY_RELAYLOG_SPILL_THRESHOLD`, (2) is greater than or equal to 33554432 bytes (32 MB), and (3) is less than or equal to 4294967296 bytes (4 GB).
+- FR16.2. When a CRST clause sets `IN_MEMORY_RELAYLOG_SPILL_THRESHOLD` alone, the server must validate that the provided value: (1) is greater than or equal to 8388608 bytes (8 MB), and (2) is less than the current `IN_MEMORY_RELAYLOG_LIMIT`.
+- FR16.3. When a CRST clause sets both variables together, the server must validate that: (1) the provided `IN_MEMORY_RELAYLOG_SPILL_THRESHOLD` is greater than or equal to 8388608 bytes (8 MB), (2) the provided `IN_MEMORY_RELAYLOG_LIMIT` is greater than or equal to 33554432 bytes (32 MB) and less than or equal to 4294967296 bytes (4 GB), and (3) the provided `IN_MEMORY_RELAYLOG_SPILL_THRESHOLD` is less than the provided `IN_MEMORY_RELAYLOG_LIMIT`.
```
karolina-szczepankiewicz, line 30: Please add the following error codes and messages to the design:

- `ER_CRST_IN_MEMORY_RELAYLOG_ONLY_FOR_CSA`
- `ER_REPLICA_IN_MEMORY_RELAYLOG_INCOMPATIBLE_CONFIGURATION`
- `ER_CRST_IN_MEMORY_RELAYLOG_INVALID_CONFIG`

Also clarify rejection timing

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+- FR30.1. On `START REPLICA IO_THREAD` after `STOP REPLICA IO_THREAD`, the receiver must automatically re-fetch the previously truncated transaction based on GTID auto-positioning, so the applier can re-dispatch and re-apply the transaction.
+- FR30.2. The queue must treat a truncated envelope as a terminal state. When a truncated envelope reaches the head, the sweep advances the commit cursor over it, removes it from the queue, and frees its reserved memory on destruction.
+- FR31. On `RESET REPLICA`, the feature must perform queue reset, and delete any spill files and the error-diagnostic file. `gtid_executed` is left unchanged, so `START REPLICA` can re-fetch by GTID auto-positioning.
+- FR32. The feature must introduce a new Performance Schema table, `replication_in_memory_relaylog`, with one row per channel. Metrics are collected in memory and materialized on read (not persisted; cumulative since replication started; reset on `RESET REPLICA`; zeroed after a server restart). For the full list of fields, refer to the High Level Design: Observability section.
```
karolina-szczepankiewicz, line 51: FR32 proposes a new `performance_schema.replication_in_memory_relaylog` table. Please consider adding the IMR metrics to `replication_applier_metrics` instead. It already provides per-channel applier metrics, and CSA can populate only the new IMR columns. If a separate table is preferred, please explain why these metrics need a separate Performance Schema table.

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+
+Trx_payload
+`Trx_payload` carries the byte stream `Fetchable_transaction` of each transaction with `RAII`(Resource Acquisition Is Initialization) memory-usage accounting. The owner of the `Fetchable_transaction` changes as the transaction moves through different phases:
+Open (IO thread still receiving): held by the receiver and the `Transaction_envelope` entry in the queue.
+Sealed, not yet dispatched (fully received, waiting in the deque): held by the `Transaction_envelope` entry only; the receiver dropped its reference after marking the envelope `sealed`.
+Dispatched, applying (in-flight): held by the `Transaction_envelope` entry and the `Job_applier` (a copy taken at dispatch).
+Committed: the `Transaction_envelope` entry resets its `Trx_payload` pointer, and the Job is destroyed. With the drop of the last reference, `Fetchable_transaction` is freed while the empty `Transaction_envelope` entry lingers behind an uncommitted head.
+Rolled back (stop or retry): the `Job_applier` is dropped by the workers but the `Transaction_envelope` entry keeps its reference, so the uncommitted payload stays alive and re-dispatch can hand it to a fresh `Job_applier`.
+
+Queue
```
karolina-szczepankiewicz, line 80: Please consider owning Trx_envelope_queue in an IMR runtime object outside Master_info. Master_info mainly represents connection and repository state, so it is not the most natural owner for a large volatile queue. For example, an IMR runtime object similar to the CSA runtime object could own the queue. Its lifetime must allow the queue to survive `STOP REPLICA SQL_THREAD` while the IO thread continues receiving. The persisted options can remain in the appropriate channel repository object.

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+  - `In_Memory_Queue_Length` — the current number of transaction envelopes held in the queue.
+
+  For a channel that does not use the in-memory relay log, both columns report 0.
+- FR35. Receiver-side (IO thread) and queue-side (in-memory queue) errors introduced by the feature must be reported through the standard receiver error channel `SHOW REPLICA STATUS` (`Last_IO_Errno` / `Last_IO_Error`/`Last_IO_Error_Timestamp`) and the server error log, exactly as IO-thread errors are reported today.
+- FR36. The memory path must populate `performance_schema.replication_applier_status_by_worker` (`LAST_ERROR_NUMBER`, `LAST_ERROR_MESSAGE`, `APPLYING_TRANSACTION`) in the same manner as the existing CSA design, so that failing transactions are identified by their GTID.
+- FR37. On apply error, the failing transaction must be retained so the transaction can be inspected with event-dump tooling. For a memory path transaction, the transaction payload must be written to an error-diagnostic file in relaylog format; a spill path transaction is already on disk and its file is renamed and kept.
```
karolina-szczepankiewicz, line 60: FR37 preserves only the failed transaction. In practice, surrounding transactions are often useful when investigating an apply failure. Please consider writing all possible currently retained queue entries, or explain why one transaction is enough. Please also state whether an incomplete last transaction is included.

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+Each channel reports the following runtime per-channel metrics through the `performance_schema` replication tables. All of these metrics are collected through the existing CSA statistics infrastructure (`Statistics_map` / `Statistics_monitor`, per channel). New keys are added and updated inline at their relevant points. No new collection framework is introduced.
+
+Memory Path:
+`memory_bytes_used` — current memory path bytes held (the accountant value).
+`memory_bytes_limit` — the configured `IN_MEMORY_RELAYLOG_LIMIT`.
+`memory_trx_count` — cumulative transactions admitted to the memory path.
+`memory_trx_bytes` — cumulative bytes of transactions admitted to the memory path.
+
+Spill Path:
+`disk_trx_count` — cumulative transactions to the spill path.
```
karolina-szczepankiewicz, line 200: Please prefix IMR metrics with `imr_`. Also use `imr_spill_*` instead of `imr_disk_*`, because these values describe the spill path, not the standard disk relay log.

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Configuration / Knobs — New configuration clauses or options
+
+All In-Memory relaylog settings are per channel. They are exposed as `CHANGE REPLICATION SOURCE TO` clauses. Each clause is a per-channel value held in the channel's in-memory `Master_info` and backed by the persisted replication metadata repository. They are effective only on a CSA channel with asynchronous replication. Setting them on any other channel type is rejected with an error.
+
+To enable In-Memory relaylog on a CSA channel:
+
+```sql
+CHANGE REPLICATION SOURCE TO
+    IN_MEMORY_RELAYLOG_ENABLED         = 1,
```
karolina-szczepankiewicz, line 129: The design documents 0 | 1 and uses IN_MEMORY_RELAYLOG_ENABLED = 1, but the implementation commit accepts only ON | OFF. Please change if applicable.

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+A failed transaction is identified by its GTID. The existing `performance_schema.replication_applier_status_by_worker` fields (`LAST_ERROR_NUMBER` / `LAST_ERROR_MESSAGE`, `APPLYING_TRANSACTION` and its timestamps) continue displaying the erroneous transaction GTID with error message.
+The feature writes the erroneous transaction to a diagnostic file in relaylog format in the dedicated directory `#in_memory_relaylog_temp_files`. The file can be inspected with existing event-dump tooling, restoring the capability to investigate the relaylog as disk path. The retained diagnostic file survives server startup, and is only removed when the replication channel restarts or the channel is reset.
+
+### User Interface
+
+### Configuration / Knobs — New configuration clauses or options
+
+All In-Memory relaylog settings are per channel. They are exposed as `CHANGE REPLICATION SOURCE TO` clauses. Each clause is a per-channel value held in the channel's in-memory `Master_info` and backed by the persisted replication metadata repository. They are effective only on a CSA channel with asynchronous replication. Setting them on any other channel type is rejected with an error.
```
karolina-szczepankiewicz, line 123: Please change `Master_info` to `Relay_log_info` in this paragraph

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+`STOP REPLICA SQL_THREAD` (applier only; receiver keeps running). The queue and all uncommitted envelopes still stay in memory. In-flight jobs are driven to a terminal state (commit, or roll back), so no transaction is left half-applied. With no advance of `commit_seqno`, all uncommitted transactions remain in the queue. The IO thread keeps enqueuing new envelopes until the memory usage counter reaches the `IN_MEMORY_RELAYLOG_LIMIT`. On next `START REPLICA SQL_THREAD`, the coordinator rewinds `dispatch_seqno` to `commit_seqno` and re-dispatches all the retained uncommitted transactions in order; committed-but-not-yet-swept envelopes are skipped because their payload is freed.
```
karolina-szczepankiewicz, line 99: The implementation performs one final sweep_committed() after the SQL thread and workers stop. Please change this paragraph if applicable.

design/replication/684-in-memory-relaylog/high-level-description.md
```diff
@@ -0,0 +1,60 @@
...
+
+In the current CSA code path (referred to as the disk path in this document), a transaction needs to go through disk I/O twice before it can be applied. The receiver (IO thread) writes encoded binlog events received over the network to relaylog files, and the applier reads them back to schedule and apply the transaction.
+
+In-memory relaylog keeps the transaction in memory instead. As a new transaction arrives, the receiver wraps the transaction metadata and incoming encoded events into a data structure, pushes it to an in-memory queue. The coordinator reads from it and dispatches the transaction to workers, without relaylog file write/read during the process (referred to as the memory path in the document).
+
+The queue has a hard memory limit per channel. A transaction larger than a configured threshold is re-routed to a temporary file in standard relaylog format (referred to as the spill path in the document). On replication stop, server restart, or crash, the queue is emptied and any uncommitted transactions are re-fetched by GTID auto-positioning.
```
karolina-szczepankiewicz, line 9: Please change ``` On replica stop ``` to ``` On full replication channel stop ```

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+
+**PRIVILEGES REQUIRED**: REPLICATION_SLAVE_ADMIN
+
+**DESCRIPTION**: Enables the in-memory relaylog for the channel, replacing the standard relaylog files on the disk path. When set to 0, the feature is off, and the channel uses the standard relaylog disk path with no change in behavior. Effective only on CSA channels; enabling it elsewhere is rejected.
+
+
+
+**NAME**: `IN_MEMORY_RELAYLOG_LIMIT`
+
+**VALUES:** unsigned integer, bytes. Range [33554432 (32 MB), 4294967296 (4 GB)]; values outside the range are rejected.
```
karolina-szczepankiewicz, line 149: Please note that this range will require a 64 bit integer. If you would like to keep a 32 bit value, please change the maximum to 4 GiB - 1.

design/replication/684-in-memory-relaylog/high-level-description.md
```diff
@@ -0,0 +1,60 @@
...
+
+With In-memory relaylog, file based relaylog-related surfaces are reduced.
+
+`SHOW REPLICA STATUS` relaylog position fields (`Relay_Log_File`, `Relay_Log_Pos`, `Relay_Log_Space`) are not applicable
+
+`sync_relay_log` is not applicable
+
+`SHOW RELAYLOG EVENTS` is not applicable
+
+`FLUSH RELAY LOGS` is a no-op.
```
karolina-szczepankiewicz, line 57: Please note that the implementation must block standard relay log rotation to keep this operation a no-op.

karolina-szczepankiewicz (commented):
The design is in good shape. I reviewed the design and used the provided code only for validation. I requested clarification where the implementation and design diverge. I recommend resolving some comments through discussion, but most can be addressed during implementation and code review. Thank you!

design/replication/684-in-memory-relaylog/requirements.md
nacarvalho: IMHO we should keep the default as OFF, given that users expect durability of relay logs by default.
User stories on https://github.com/mysql/mysql-server/pull/743/changes/667fcff05a23f12390bcb0f92cb764aa8e170fbb#diff-8bcbe8983b9d223e1cb768c4a672500fb8844b4b9e8638d6d8fb1023d6fef248 also need to be updated.

design/replication/684-in-memory-relaylog/high-level-design.md
```diff
@@ -0,0 +1,224 @@
...
+`dispatch_seqno`: the next envelope to dispatch by the coordinator.
+Queue admission of `Transaction_envelope` is decided at the GTID event using `trx_length`:
+
+```cpp
+trx_length > disk_path_threshold      -> spill path
+queue_bytes + trx_length <= limit     -> memory path
+otherwise                             -> block until commits free space, then memory path
+```
+
+Reserving the exact `trx_length` up front guarantees a memory-path transaction fits before streaming begins. A transaction larger than the `IN_MEMORY_RELAYLOG_SPILL_THRESHOLD` always takes the spill path.
```
nacarvalho, line 53: It would be good to list the reasons to have both IN_MEMORY_RELAYLOG_SPILL_THRESHOLD and IN_MEMORY_RELAYLOG_LIMIT.

nacarvalho (commented):
Thank you for the contribution.
I only looked into the design, please see my comments inline.

design/replication/684-in-memory-relaylog/high-level-description.md
```diff
@@ -0,0 +1,60 @@
...
+
+Provide per-channel `CHANGE REPLICATION SOURCE TO`(CRST) knobs to enable the feature and adjust the memory limit and spill path threshold.
+
+Expose per-channel observability metrics for memory path usage, spill path usage, and receiver-side block.
+
+Report apply failures through GTID-based diagnostics `performance_schema.replication_applier_status_by_worker`, and write the failed transaction to a diagnostic file in relaylog format for easy inspection.
+
+**Out of Scope / Limitations**
+
+The feature is effective only on CSA channels. It is not supported for `Group Replication` channels, or `Semisynchronous` replication. Enabling this feature on such channels shall be rejected.
```
nacarvalho, line 43: Need to be clarified that for Group Replication, only group_replication_applier channel cannot enable it.
group_replication_recovery channel can enable it, same way it can enable CSA.
This also impacts FR2.

nacarvalho (commented):
Pending comment.

design/replication/684-in-memory-relaylog/requirements.md
henrikingo-oracle: I'm still catching up, but isn't CSA itself a new thing that was just released? If so, I think we could assume most users didn't try it yet and those who did can tolerate if a relay log by default is written or not.

That doesn't mean there couldn't be other reasons why it is good to default to OFF anyway. (Such as reduced durability in disaster situations?) But I wouldn't automatically just lean on "users expect" when CSA itself is a new way to do replication. We can in that case (re)define what the users should expect here.

If this is also added to some form of replication that was used before 2026, I would agree that relay logs files simply disappearing is not good, the default should in such cases be the old behavior.

sql/changestreams/apply/storage/relay_log/cached_event_payload.cpp
yangxt-aws: Yes, I will submit a separate PR to address the `Cached_event_payload` issue.

design/replication/684-in-memory-relaylog/high-level-design.md
yangxt-aws: This is a great question, based on my understanding `Clock_lwm_registry` is a used by applier thread to track LWM.

The `commit_seqno` is specifically utilized by the queue implementation (the deque). The main purpose is to track the head index of the deque, so the `dispatch_seqno` and `insert_seqno` can be fast accessed through index by subtracting `commit_seqno`. It does hold similar meaning as `Clock_lwm_registry`.
The `commit_seqno` maintains a different life cycle from `Clock_lwm_registry`, and its position marks where Txns may need reassign during recover (when `stop replica sql_thread` issued and half executed Trx aborted.)

design/replication/684-in-memory-relaylog/requirements.md
yangxt-aws: My original intention is to keep it ON by default. The implementation is a gap. I think it would be a good option to keep it `ON` for innovation release, and set to `OFF` for LTS.

design/replication/684-in-memory-relaylog/high-level-design.md
yangxt-aws: > After the latest changes, my understanding is that workers do not access the queue structure or its cursors. On commit, a worker only updates the dispatched envelope's completion state and releases its payload. The coordinator later advances commit_seqno and removes entries.

Exactly as you described. The worker would not touch the queue structure.

I will update the text to make it more unambiguous.

design/replication/684-in-memory-relaylog/requirements.md
henrikingo-oracle: Hmm... Maybe that's something we could use an Early Access release for (essentially alpha, feature preview releases) but I'm not sure I'm comfortable with intentionally turning it on and off in innovation releases. I see Innovation releases as dress rehearsals for LTS, so the last innovation release should essentially be very close to what then becomes LTS after a few months and there shouldn't be last minute changes just because the next release is LTS, quite the opposite in fact.

nacarvalho: We should go with OFF or ON independently of the version, Henrik mention of surprises is the main reason.

Historically we usually added new features disabled by default, especially when they imply new behaviour.
Given the in-memory relay log feature is implemented on top of CSA which has released on  26.7.0, we can make in-memory relay log ON by default given it will only be considered when CSA is used.
That is, users that do upgrade and continue to use the old applier will retain their behaviour.

yangxt-aws: @nacarvalho @henrikingo-oracle Thank you so much for explaining my gap of understanding. I also think it's a good option to have it default ON with CSA, while the default MTA will continue function in current style. I will address the code gap and update the user story in the next revision.

design/replication/684-in-memory-relaylog/high-level-design.md
yangxt-aws: @karolina-szczepankiewicz This is a very valuable suggestion and provide me a lot inspiration. I think having an IMR runtime object would be a good improvment.

The part I am hesitant is the lifetime mismatch. In CSA, per-channel object is created in the stack of `csa_service::run()`, and they are cleanly destroyed when SQL thread terminates.

However, an IMR trx queue is designed to outlive both IO and SQL threads. When owned by the IMR runtime object, the queue's destruction would be decoupled from channel object, and rely on the channel termination notification from different places, which I think would bring in extra leak risk.

I am leaning towards having an IMR runtime object acting as a stateless factory, that creates the IMR queue, and hands the ownership to `Relay_log_info`. This way, the queue dies with the channel object, and it also conform with the on-disk relay_log ownership.

karolina-szczepankiewicz: Just to clarify, I was referring to the ownership pattern of Csa_service. It lives for the lifetime of the server and holds per-channel state. Its active channel entries currently follow the SQL thread lifetime, but a similar IMR object could keep its entries for the lifetime of the replication channel.

Of course, Csa_service would still access IMR through a non-owning pointer. My suggestion is to keep IMR ownership and lifetime separate, with similar bookkeeping in its own object.

Relay_log_info already has many responsibilities, and separating them becomes harder as more are added over time. Master_info has a somewhat clearer scope. This is why I suggested keeping the IMR runtime state in a separate object, but I leave the final design and ownership decision to you :)

design/replication/684-in-memory-relaylog/high-level-design.md
karolina-szczepankiewicz: Thank you, that clarifies it. We can mark this as resolved.

design/replication/684-in-memory-relaylog/high-level-description.md
```diff
@@ -0,0 +1,60 @@
...
+
+`SHOW REPLICA STATUS` relaylog position fields (`Relay_Log_File`, `Relay_Log_Pos`, `Relay_Log_Space`) are not applicable
+
+`sync_relay_log` is not applicable
+
+`SHOW RELAYLOG EVENTS` is not applicable
+
+`FLUSH RELAY LOGS` is a no-op.
+
+Because in-flight transactions are memory-only, replication restart (`STOP REPLICA` then `START REPLICA`) may need to re-fetch more transactions from the source than a file-based relaylog replica.
```
tiagoportelajorge, line 59: I think that this needs to be more assertive. It actually needs to refetch all missing transactions. The comparison is not needed.

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
+### Functional Requirements
+
+- FR1. The feature must provide a per-channel option, introduced as a new `CHANGE REPLICATION SOURCE TO` option `IN_MEMORY_RELAYLOG_ENABLED`, that controls whether the feature is enabled. It is `ON` by default for eligible channels. When `OFF`, replication must fall back to the standard disk path relaylog.
+- FR2. The feature must only apply to CSA channels with asynchronous replication. Enabling the feature on any unsupported channel type must be rejected with a clear error.
+- FR3. The feature must store each event in the same encoded bytes as in disk path relaylog, so that the existing decode and apply path can continue working without modification.
+- FR4. The receiver must create an object of type `Transaction_envelope` when it encounters a new transaction. The receiver appends the transaction payload to the object as it processes incoming events belonging to the same transaction.
```
tiagoportelajorge, line 6: Given that FRs should be testable and added to a test matrix so that a QA Engineer can create a test to validate the behaviour, I don't think FR4 and its children belong here. These are low-level implementation details.

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+- FR4.2. Upon receiving the GTID event, the receiver must select the store path for the payload by comparing its `trx_length` with configured `IN_MEMORY_RELAYLOG_SPILL_THRESHOLD`. A transaction with a `trx_length` that exceeds the threshold takes the spill path. Otherwise, it takes the memory path.
+- FR4.3. As the receiver processes each incoming event in a transaction's payload, it must advance the transaction end position to notify the worker. This allows a worker to begin applying already received events while the rest of the transaction is still being received.
+- FR5. The feature must not change receiver side GTID tracking. A GTID must be added to `Retrieved_Gtid_Set` only after the transaction has been marked as sealed.
+- FR6. The coordinator must build and dispatch a `Job_applier` from the `Transaction_envelope` object, and workers must decode and apply the transaction in the same manner as the existing code path.
```
tiagoportelajorge, line 11: Given that FRs should be testable and added to a test matrix so that a QA Engineer can create a test to validate the behaviour, I don't think FR6 and its children belong here. These are low-level implementation details.

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+- FR8.1. The counter is an atomic variable. It increases by the transaction's size when a memory path payload is created, and decreases by the same amount when that payload is released at commit.
+- FR9. The queue must be tracked by three cursors: (1) A commit cursor that marks the last contiguously committed transaction, meaning all transactions before this cursor have been committed with no gaps. (2) A dispatch cursor that marks the next available transaction envelope to be dispatched. (3) An insert cursor that marks the position where the next incoming envelope should be appended.
+- FR9.1 The cursors must strictly follow the invariant: `commit cursor` <= `dispatch cursor` <= `insert cursor`.
+- FR10. The queue must support concurrent access by the receiver, enqueueing at the tail, and by the coordinator and workers, reading from the head. Each transaction envelope is retained from the time its GTID event is received and dequeued after its transaction commits.
+- FR10.1. The queue must be protected by a queue-level mutex that guards its structure. All operations that mutate the queue structure and the cursors must hold this mutex.
```
tiagoportelajorge, line 21: Have you considered something in the line of https://apoorvsachan.com/blog/lockfree_atomic_queues_with_cas/

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+- FR18.1. If the `#in_memory_relaylog_temp_files` directory already exists at startup, the server must clean up its managed temp files according to FR25. This is safe because the channel invalidates and discards all previously fetched relaylog content and recovers solely by GTID auto-positioning based on `gtid_executed`.
+- FR18.2. At startup, if the `#in_memory_relaylog_temp_files` path exists but is not a directory (e.g., a symlink or regular file), or if directory creation or permission acquisition fails, the feature must log a clear error message and reject starting replication channels with `IN_MEMORY_RELAYLOG_ENABLED` set to ON.
+- FR19. The feature must exclude the `#in_memory_relaylog_temp_files` directory from schema-visible listings such as `SHOW DATABASES` and information_schema, so that it does not appear as a schema.
+- FR20. A channel's spill path files count against `relay_log_space_limit` as a soft limit. The receiver finishes writing the current transaction's spill file even if the total disk usage exceeds the limit, then blocks the next spill path transaction until committed files are deleted and the total size drops back under the limit.
```
tiagoportelajorge, line 36: Can't you know up hand to it?

design/replication/684-in-memory-relaylog/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+- FR21. The spill path file store must use exactly one self-contained file per transaction with a unique filename to avoid conflicts. The file must be deleted automatically once its transaction commits.
+- FR22. A spill file must be created with the same file permissions, ownership, and encryption as the channel's relaylog files, since it contains equally sensitive data and resides alongside the relay log.
+- FR23. If disk space is exhausted while writing to a temp file during transaction execution, the write error must be handled identically to a disk-full error on the standard binlog cache temp file.
+- FR24. A spill path temp file created by the feature must have a name matching the pattern `imr_sp_<channel>_<unique_id>`, where `<channel>` is a lowercase channel identifier, and `<unique_id>` is a lowercase identifier unique within `#in_memory_relaylog_temp_files` (IMR SP stands for "In-Memory Relaylog Spill Path").
+- FR25. Startup cleanup must delete only spill files whose basename matches the temp file naming pattern `imr_sp_*` and must reject all other directory entries.
```
tiagoportelajorge, line 41: Will you clean up even if you disable the feature?

Imagine the case where you enable the feature, have a lot of spill files, and then you shut down and decide to change back to the old behaviour. Will the old spilled files remain in place?

yangxt-aws: Yes, all the spill files on the disk would be removed during mode change from in-momery on -> off.
Technically, when `stop replica` issued, the queue is emptied and leftover disk spill files would be cleaned up.

design/replication/684-in-memory-relaylog/requirements.md
yangxt-aws: Thank you for the link and idea. This is a very helpful and inspiring suggestion. I did try out lockfree queue implementation using circular buffer + atomic cursors. The overall benchmark throughput is basically the same as a deque + mutex.
Here's my two cents:
1. The major bottleneck on applier thread is worker applying the jobs. (where the lock is needed when dequeue an entry)
2. The major bottleneck on receive thread is appending events to an existing entry in the queue. (where the lock is needed when inserting new group)

Thus removing the lock does not improve the overall performance in a measurable way.

I therefore change the implementation back to deque + mutex, for easier thread handling, and simpler correctness guarantee.

design/replication/684-in-memory-relaylog/requirements.md
yangxt-aws: Yes, this is to mirror the behavior of traditional in-disk relaylog behavior that it rotates the log file after it surpass the set limit.

sql/changestreams/apply/storage/in_memory/event_set_fetchable_memory.cc
```diff
@@ -0,0 +1,279 @@
...
+    m_stream_sealed = true;
+  }
+  m_stream_cv.notify_one();
+}
+
+void Event_set_fetchable_memory::set_stream_truncated() {
+  // Mark the owning transaction truncated, then this batch.
+  if (m_fetchable_trx != nullptr) m_fetchable_trx->set_fetching_truncated();
+  // Mark the owning envelope truncated.
+  if (m_envelope != nullptr) m_envelope->set_truncated();
```
tiagoportelajorge: What keeps this object alive after set_truncated()? Once the envelope is marked truncated, the coordinator can skip it and remove it from the queue. If no job has picked it up yet, that also destroys the event source, while the receiver still has the mutex access and notification below to execute. The receiver only holds a raw pointer. I think it needs to keep ownership until it has finished with the stream. The spill implementation looks to have the same issue.

sql/changestreams/apply/storage/in_memory/queued_transaction_writer.cc
```diff
@@ -0,0 +1,73 @@
...
+
+bool append_transaction_event(Streaming_event_sink *&current_sink,
+                              const char *buf, std::size_t len,
+                              bool is_terminal) {
+  // Defensive: no group is open, so there is nothing to append into.
+  if (current_sink == nullptr) return true;
+
+  // Forward the raw encoded bytes straight to the sink.
+  // On the terminal event, the append also seals the stream.
+  current_sink->append_event(buf, len, /*seal_after=*/is_terminal);
```
tiagoportelajorge: How does the receiver find out that a spill write failed? append_event() returns nothing, and this function returns success even if opening, writing or flushing the file failed. If this is the last event, we also clear the sink and let queue_event() add the GTID to Retrieved_Gtid_Set. We would then be saying that we received a transaction whose bytes we failed to store. The error needs to come back through this call before we update the completed-transaction bookkeeping.

sql/changestreams/apply/storage/in_memory/queued_transaction_reader.cc
```diff
@@ -0,0 +1,112 @@
...
+  return job;
+}
+
+bool Queued_transaction_reader::is_stopped() const {
+  return m_is_error || m_stopped || m_queue->is_stopped();
+}
+
+bool Queued_transaction_reader::is_error() const { return m_is_error; }
+
+void Queued_transaction_reader::stop() {
```
tiagoportelajorge: This wakes the coordinator, but what happens to a worker waiting for the next event in a partially received transaction? The memory and spill readers wait on different condition variables, and neither checks whether SQL is stopping. If the source stops sending, STOP REPLICA SQL_THREAD can get stuck there. Full STOP also waits for SQL before stopping IO, so we cannot rely on IO shutdown to wake the worker. We need to interrupt the read without throwing away the transaction, since SQL-only restart must be able to replay it.

sql/rpl_replica.cc
```diff
@@ -11728,6 +12144,19 @@ static void check_replica_configuration_restrictions() {
 */
 static bool check_replica_configuration_errors(Master_info *mi,
                                                int thread_mask) {
+  // The in-memory relay log is incompatible with semi-synchronous replication.
+  // Only the receiver (IO thread) connects to the source and acknowledges, so
+  // the gate applies to an IO-thread start on an in-memory channel. This is a
+  // start-time (not CHANGE-time) check because semisync is a plugin whose
+  // enablement is not known when CHANGE REPLICATION SOURCE runs. Leaves the
+  // queue intact -- it gates running the receiver, not the queue's existence.
+  if ((thread_mask & REPLICA_IO) && mi->is_in_memory_relaylog() &&
```
tiagoportelajorge: This check can be bypassed by starting with semisync OFF, turning it ON while IO is running, and then reconnecting. The plugin negotiates semisync when requesting the dump again, without coming back through this START check. We could then send acknowledgements for events that are only in memory. The restriction needs to hold on reconnect too, or we need to reject enabling semisync while an IMR receiver is running. That sequence is worth adding to the test.

sql/changestreams/apply/storage/in_memory/event_set_fetchable_spill.cc
```diff
@@ -0,0 +1,359 @@
...
+  {
+    std::lock_guard<std::mutex> lock(m_stream_mutex);
+    if (is_error() || !m_stream_open || m_stream_sealed || m_stream_truncated) {
+      return;
+    }
+  }
+
+  // Write the raw bytes and flush so a concurrent reader can see them.
+  if (m_writer->append_raw(buf, len) || m_writer->flush()) {
+    m_failure_msg.assign(m_writer->get_error_str());
```
tiagoportelajorge: Aren’t these fields shared with the worker? The receiver writes m_failure_msg and m_status here, while the worker reads them without taking the same lock. That gives us a data race when a write or flush fails. We need consistent locking around the error state; returning a reference to the string is also unsafe if another thread can change it. The successful streaming tests won’t catch this—we need a concurrent test that actually makes an IO operation fail.

sql/sql_lex.h
```diff
@@ -481,6 +492,19 @@ struct LEX_SOURCE_INFO {
   /// The maximum amout of memory that can be used by the channel to keep
   /// binlog events
   ulong applier_event_memory_limit{applier_event_memory_limit_unspecified};
+  /// constant - unspecified IN_MEMORY_RELAYLOG_LIMIT option
+  static constexpr int in_memory_relaylog_limit_unspecified{0};
```
tiagoportelajorge: Using zero for “not specified” means we also ignore an explicit IN_MEMORY_RELAYLOG_LIMIT = 0, rather than rejecting it. The spill threshold has the same problem. This is separate from the 4 GiB issue already discussed. We need to distinguish an omitted clause from a supplied value of zero. Please add both zero cases to the validation test; checking just below the minimum doesn’t catch this.

sql/rpl_replica.cc
```diff
@@ -2189,10 +2266,17 @@ bool start_slave_threads(bool need_lock_slave, bool wait_for_start,
...
       }
     }
-    if (!is_error)
+    if (!is_error) {
+      // Arm the applier role before creating the SQL (coordinator) thread, so
+      // dispatch_next() serves it (the queue rests stopped). Caller-thread,
+      // under the channel start/stop serialization; the SQL thread never
+      // creates the queue.
+      if (mi->is_in_memory_relaylog())
+        mi->m_trx_queue->resume(mysql::csa::Trx_envelope_queue::Scope::APPLIER);
```
tiagoportelajorge: If SQL-thread creation fails here, who puts the queue’s applier role back into the stopped state? We resume it before creating the thread, but the failure cleanup below only stops IO and the monitor. The queue can therefore think SQL is still running and skip full-stop cleanup. IO might already have queued transactions by then. We should undo the role change on startup failure and check the corresponding IO-start case as well.

sql/changestreams/apply/storage/in_memory/event_set_fetchable_memory.cc
```diff
@@ -0,0 +1,279 @@
...
+    {
+      std::lock_guard<std::mutex> lock(m_stream_mutex);
+      if (m_event_id >= m_events.size()) {
+        return {};
+      }
+      reader_event = m_events[m_event_id++];
+      is_last_in_batch = m_stream_sealed && m_event_id == m_events.size();
+    }
+
+    auto current_event = reader_event->decode();
```
tiagoportelajorge: decode() can return an empty pointer, but we dereference it on the next line. The new decoder allocates a copy of the event and explicitly returns empty if that allocation fails. We should report an applier read error here instead of crashing. I noticed the older cache reader has a similar unchecked call, but this new allocation gives us a concrete failure case to handle and test.

design/replication/684-in-memory-relaylog/requirements.md
tiagoportelajorge: Thanks, that covers a normal STOP. What about a crash, though? The destructors won’t run, so spill files can be left behind. If the server then starts with IMR disabled, what removes them? There won’t be an old queue to destroy at that point. I think we should spell out that case and test it

design/replication/684-in-memory-relaylog/requirements.md
karolina-szczepankiewicz: The linked queue implementation does not provide a blocking wait when empty. We need the coordinator to sleep when replication is idle, so it is not a direct replacement for the current queue.

A mutex shared by all workers can become a single contention point and limit scaling as the number of workers grows. This is why the important parts are: keeping workers from acquiring the queue mutex, and lazy removal of entries by the coordinator, which is already covered by the current design.

Please ensure that workers do not acquire the queue mutex while applying or committing transactions, including payload release. This restriction does not apply to error handling or shutdown.

PR #774, opened 2026-09-21
Bug#121332 Empty --datadir= value asserts mysqld in debug builds

VarunNagaraju:
https://bugs.mysql.com/bug.php?id=121332

Problem:
=======
Starting a debug build with an empty --datadir value aborts on an assertion during early startup, before anything reaches the error log:

  mysqld: sql/mysqld.cc: bool initialize_manifest_file_components():
  Assertion `strlen(mysql_real_data_home) > 0' failed.

main.keyring_migration_password reaches this from the test suite itself and fails with:

`Pattern "\[ERROR\] \[MY-[0-9]+] \[Server\] Aborting" not found`

Analysis:
======
This is a regression from the fix for Bug#39181231 (e2863e9). Plugin_and_data_dir_option_parser::Plugin_and_data_dir_option_parser() now does memset(mysql_real_data_home, 0, sizeof(...)) followed by strncpy() of the option value. Previously it used
memcpy(dst, src, strlen(src)), which for an empty value copied zero bytes and left the compiled-in default ("data/", installed earlier in mysqld_main() by
strmake(mysql_real_data_home, get_relative_path(MYSQL_DATADIR), ...)) in place. The new unconditional memset clears that default, and strncpy of "" leaves the buffer empty.

The existing guard

`  if (!datadir) { ... /* keep the current value */ ... }`

covers a missing option only. --datadir is GET_STR/OPT_ARG, so --datadir= yields a non-NULL pointer to an empty string and the guard is bypassed.

--plugin-dir= has the same gap in the same constructor. It is harmless today only because plugindir_ falls back to
get_relative_path(PLUGINDIR) later.

main.keyring_migration_password spawns mysqld with --datadir=$MYSQLD_DATADIR, but $MYSQLD_DATADIR is not set by mysql-test-run.pl and the test never defines it - the other tests using that name each set it themselves with
"let $MYSQLD_DATADIR= `SELECT @@datadir`". The option therefore expands to a literal --datadir= . Before e2863e9 the zero-byte memcpy left the compiled-in default alone, so the server still got as far as failing keyring migration and logging "Aborting", and the test passed by accident.

Fix:
===
Treat an empty option value the same way as a missing one, for both --datadir and --plugin-dir.

Test:
====
Stop main.keyring_migration_password depending on the undefined variable. It now passes a private datadir under $MYSQL_TMP_DIR that the test creates and removes; defining $MYSQLD_DATADIR "correctly" instead would have pointed the spawned instance at the datadir of the server MTR is already running.
--no-defaults is passed first so that a system or user my.cnf cannot change the options under test.

The generic "Aborting" match is replaced by the specific message chain the clean exit produces - keyring migration failure, Aborting, and Shutdown complete - so that a future silent exit before the abort cannot pass the test again.

### How was it tested?

- [ ] Added/updated MTR tests under `mysql-test/`
- [x] Ran the relevant full suite (name it): main

### Contributor checklist

### AI assistance


VarunNagaraju:

PR #776, opened 2026-09-22
Bug#121262 `SET SESSION foreign_key_checks = DEFAULT` incorrectly sets the session value to `0`

arthurschreiber:
### What does this change do?

Fixes `SET SESSION <var> = DEFAULT` for the two bit-backed system variables declared with `REVERSE()` semantics, `foreign_key_checks` and `unique_checks`, which were left with the opposite of the global value. BUG#121262.

### Why is it needed?

`SET SESSION <var> = DEFAULT` does not correctly reset the values of the `foreign_key_checks` and the `unique_checks` variables:

```sql
SELECT @@foreign_key_checks;                 -- 1
SET @@foreign_key_checks = 0;
SET SESSION foreign_key_checks = DEFAULT;
SELECT @@foreign_key_checks;                 -- actual: 0   expected: 1
```

Code that relies on `= DEFAULT` to restore session variables (e.g. a MySQL connection pools) will silently run with the wrong session variables. Especially with `foreign_key_checks` and `unique_checks` being disabled when it is expected to be enabled, this can lead to data silent and hard to understand data inconsistency.

### How was it tested?

- [x] Ran the relevant full suite (name it): ______

### Contributor checklist

### AI assistance

- [x] I did not use AI assistance for this contribution
- [ ] I used AI assistance for this contribution

Claude Code (Fable 5.1) was used for test generation and self-review.

### Areas touched

`sql` (system variables, `Sys_var_bit`), `mysql-test` (`main`, `sys_vars`, `binlog` suites), replication-visible via `Query_log_event` `flags2`.


arthurschreiber:
The "Format Check" CI failure seems to be pre-existing and unrelated to the changes in this PR:

```
sql/sys_vars.h:1513:37: error: code should be clang-formatted [-Wclang-format-violations]
#define MAX_SET(X) ((((1ULL << ((X)-1)) - 1) << 1) | 1)
```

I guess it's triggered because this PR modifies the `sys_vars.h` file?

RidhaOracle:
Yep correct the Format Check is a noisy signal, there is a conflict of clang version in our repo (15) vs what I have put in the pipeline (18). I will release a fix for this, this week.

